PR #12, opened 2015-10-05
Uses fixed version of node-crypt3

rolandpoulter:
I submitted a PR against node-crypt3, but it looks like the project is dead. I could push it to NPM under a different name?

Also the tests for this in `spec/lib/services/encryption-spec.js` do not pass on mac before or after this change.

https://hwjiraprd01.corp.emc.com/browse/MON-542

I setup travis-ci for node-crypt3 that builds it for node 0.10, 0.12, 4.0, 4.1 on linux and mac:
https://travis-ci.org/rolandpoulter/node-crypt3

### DIFF
--- a/lib/common/encryption.js
+++ b/lib/common/encryption.js
@@ -85,4 +85,3 @@ function encryptionFactory(
 
     return Encryption;
 }
-
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "blocked": "^1.1.0",
     "bluebird": "^2.8.0",
     "colors": "^1.0.3",
+    "crypt3": "git+http://github.com/rolandpoulter/node-crypt3",
     "di": "git+https://5f5fdee3b5f3bc054cece98ad854c3090043f3c0:x-oauth-basic@github.com/RackHD/di.js.git",
     "ejs": "^2.0.8",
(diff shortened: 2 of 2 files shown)


benbp:
So it's just broken on node 4 right now? I'm leaning towards just closing this rather than introduce Roland's fork as a dependency, and solving the problem when we transition to node 4.

jfrey:
We should be able to remove the dependency and use built in crypto as well.

Some time needs to be put into it though.

On Nov 3, 2015, at 10:52 AM, Ben Broderick Phillips <notifications@github.com<mailto:notifications@github.com>> wrote:


So it's just broken on node 4 right now? I'm leaning towards just closing this rather than introduce Roland's fork as a dependency, and solving the problem when we transition to node 4.

benbp:
Ok, so does that mean I should close this, or is there a reason to keep it open?

yyscamper:
If https://hwjiraprd01.corp.emc.com/browse/MAG-248 is resolved, close this PR is OK. Now I still cannot find the solution for MAG-248. Of course, if we have plenty of time before using Node 4.0, close this PR is OK as well, because I agree with @jfrey , use built-in library should be the target.

PR #26, opened 2015-11-06
Fixes for using node version 4.

rolandpoulter:


lib/common/encryption.js
```diff
@@ -78,11 +78,31 @@ function encryptionFactory(
...
+     * ['sha512', 'sha256','md5'], the default is 'sha512'
+     * @param {String} [salt] - The salt feed to hash algorithm, if not specified, the function will
+     * generate a random salt.
+     * @return {String} the salt
+     */
+    Encryption.prototype.createSalt = function (algorithm, salt) {
         algorithm = algorithm || 'sha512';
-        salt = salt || crypt3.createSalt(algorithm);
-        return crypt3(data, salt);
+        var signitures = {
```
benbp: `signatures` is the correct spelling

lib/common/encryption.js
```diff
@@ -78,11 +78,31 @@ function encryptionFactory(
...
+          md5: '$1$',
```
benbp: 4 space indentation plz :P

benbp:
Are there unit tests that need updating as well?

benbp:
Thanks for fixing!

rolandpoulter:
@benbp don't think I need to add more unit tests. I could for `createSalt` but `createHash` already has tests. I will if requested.

benbp:
Oh I guess I meant did we need to update existing ones? If not, great. :+1: when the build is passing.

jfrey:
+1

yyscamper:
Even it has been merged, I still want to raise a :+1: , glad we have move to Node 4 ,thank for your wok @rolandpoulter

PR #28, opened 2015-11-06
Attempt to make spec/helper.js more robust.

rolandpoulter:


spec/helper.js
```diff
@@ -217,11 +217,12 @@ global.helper = {
      * @returns {Promise}
      */
     stop: function () {
-        if (this.core) {
-            return this.core.stop();
-        } else {
-            return bluebird.resolve();
-        }
+      var promises = [new Promise(function (resolve) {
+        setTimeout(resolve);
```
benbp: Why is this needed?

spec/helper.js
```diff
@@ -217,11 +217,12 @@ global.helper = {
...
+      })];
+      if (this.core) promises.push(this.core.stop());
+      promises.push(bluebird.resolve());
+      Promise.all(promises);
```
benbp: Need to return.

spec/helper.js
```diff
@@ -217,11 +217,12 @@ global.helper = {
...
-        if (this.core) {
-            return this.core.stop();
-        } else {
-            return bluebird.resolve();
-        }
+      var promises = [new Promise(function (resolve) {
+        setTimeout(resolve);
+      })];
+      if (this.core) promises.push(this.core.stop());
+      promises.push(bluebird.resolve());
```
benbp: Can just do Promise.resolve() instead since you already have it in scope.

spec/helper.js
```diff
@@ -217,11 +217,12 @@ global.helper = {
...
     stop: function () {
-        if (this.core) {
-            return this.core.stop();
-        } else {
-            return bluebird.resolve();
-        }
+      var promises = [new Promise(function (resolve) {
+        setTimeout(resolve);
+      })];
+      if (this.core) promises.push(this.core.stop());
```
benbp: I think we shy away from one line ifs in our general, unwritten style conventions.

jfrey:
Need a description.

spec/helper.js
rolandpoulter: I was hoping that having a timeout between tests would yield to `nextTick` and `setImmediate` to be called. Incase it is causing race conditions.

rolandpoulter:
Closing for now, I meant to open this PR against my fork.

benbp:
lol

PR #30, opened 2015-11-09
Remove sudo and package dependencies from travis file (so we can have…

benbp:
… travis containers)

Our unit tests here shouldn't require sudo or apt updates, so remove them in order to support new travis containerized infrastructure.

Putting up a PR to test this, but I don't think it will work with on-core yet.


benbp:
Yep, the gc-stats dependency prevents us from testing without g++ :(

PR #38, opened 2015-11-30
Remove gs-stats to speed up npm install

rolandpoulter:
Installing `gc-stats` takes a long time because it has to do a lot of native compilation. In order to make npm install for `on-core` and its dependents I propose removing it.

~~I think `statsd` was only being used to broadcast info from `gc-stats` and AMQP request times. So I removed that as well. I can add `statsd` back if this is needed for something.~~

@benbp @DavidjohnBlodgett @VulpesArtificem

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,6 @@ module.exports = function (di, directory) {
                 helper.requireWrapper('pluralize', 'pluralize'),
                 helper.requireWrapper('blocked', 'blocked'),
                 helper.requireWrapper('always-tail', 'Tail'),
-                helper.requireWrapper('gc-stats', 'gc-stats'),
                 helper.requireWrapper('flat', 'flat'),
 
                 // Glob Requirables
--- a/lib/services/error-publisher.js
+++ b/lib/services/error-publisher.js
@@ -107,4 +107,3 @@ function ErrorPublisherFactory(
 
     return new ErrorPublisher();
 }
-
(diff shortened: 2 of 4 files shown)


lib/common/messenger.js
```diff
@@ -35,7 +34,6 @@ function messengerFactory(
     Serializable,
     shortid,
     Timer,
```
rolandpoulter: I think I can remove Timer too now.

lib/common/messenger.js
```diff
@@ -277,15 +275,6 @@ function messengerFactory(
                         subscription.dispose();
                     }
 
-                    statsd.timing(
-                        statsd.sanitize(
```
rolandpoulter: This is the only place where statsd is used inside `on-core`.

benbp:
I approve of this, but I'm also wondering whether we want to keep the statsd service in even thought it's not being used, since it's a good example of *how* to integrate with on-statsd?

rolandpoulter:
@benbp I think so actually. I'm going to add it back in.

DavidjohnBlodgett:
was looking over @rolandpoulter shoulder, so I approve! :+1:

rolandpoulter:
Takes about 20 seconds to npm install now.

heckj:
OMG THANK YOU

benbp:
lol

benbp:
:+1:

benbp:
I'll take that as a :+1: @heckj

PR #39, opened 2015-11-30
Update travis config to remove sudo and g++

rolandpoulter:
This was needed for `gc-stats`. Now that it is gone we can run on travis without sudo.


heckj:
+1

benbp:
:+1:

benbp:
:+1:

PR #43, opened 2015-12-09
Allow filtering of log messages based on severity level

dalebremner:
@benbp @jlongever @brianparry

### DIFF
--- a/lib/common/logger.js
+++ b/lib/common/logger.js
@@ -11,7 +11,8 @@ loggerFactory.$inject = [
     'Assert',
     '_',
     'Util',
-    'stack-trace'
+    'stack-trace',
+    'nconf'
 ];
 
 /**
@@ -24,7 +25,8 @@ function loggerFactory(
     assert,
     _,
     util,
-    stack
+    stack,
+    nconf
 ) {
     var levels = _.keys(Constants.Logging.Levels);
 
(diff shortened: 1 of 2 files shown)


spec/lib/common/logger-spec.js
```diff
@@ -6,13 +6,14 @@
...
 describe('Logger', function () {
-    var Logger, events;
+    var Logger, events, configuration;
 
     helper.before();
 
     before(function () {
         Logger = helper.injector.get('Logger');
         events = helper.injector.get('Events');
+        configuration = helper.injector.get('Services.Configuration');
```
benbp: Ignoreable.

lib/common/logger.js
```diff
@@ -57,6 +59,13 @@ function loggerFactory(
         assert.isIn(level, levels);
         assert.string(message, 'message');
 
+        // Exit if the log level of this message is less than the minimum log level.
+        var minLogLevel = nconf.get('minLogLevel');
+        if ((minLogLevel === undefined) || (typeof minLogLevel !== 'number')) {
+            minLogLevel = 0;
+        }
+        if (Constants.Logging.Levels[level] < minLogLevel) return;
```
benbp: We try to avoid braceless if statements as a style rule, if you wouldn't mind changing this.

benbp:
:+1: thanks!

iceiilin:
May I ask is there any related PRs in other repos? Or how will this filter be used?

yyscamper:
:+1:

dalebremner:
Hi Lyne. You will be able to limit the verbosity of log messages emitted by monorail by specifying the minimum level in monorail.json

heckj:
@dalebremner quick question - if I wanted to add a stanza to filter on warning of worse, then I'd need to add 

    minLogLevel: 4

to the monorail.json configuration file? Adding quick docs in https://github.com/RackHD/docs/pull/212 and wanted to verify I had it correct

benbp:
Ah I hadn't thought about that. Would be good if we could do conversion so once could specify by log level name rather than just number.

dalebremner:
@heckj correct, 4 would be "warning" or worse...

PR #45, opened 2015-12-10
Quieter normal eventloop

heckj:
Shifting logging level for event loop blocks under 50ms (from 100ms) from Notice to Debug level logging, to work with update to enable log filtering, and https://github.com/RackHD/RackHD/pull/30 to quiet down the demonstration log output


brianparry:
:+1:

benbp:
:+1:

PR #47, opened 2015-12-14
Fix bug with calling waterline create with a buffer

benbp:
Fixes https://github.com/RackHD/RackHD/issues/40

@heckj @kacole2

### DIFF
--- a/lib/common/template.js
+++ b/lib/common/template.js
@@ -41,7 +41,7 @@ function templateServiceFactory(FileLoader, Constants, Promise, _, waterline, Lo
             Constants.Templates.Directory
         ).then(function (templates) {
             var promises = _.map(templates, function (contents, name) {
-                return self.put(name, contents);
+                return self.put(name, contents.toString());
             });
 
             return Promise.all(promises);
--- a/lib/common/profile.js
+++ b/lib/common/profile.js
@@ -40,7 +40,7 @@ function profileServiceFactory(FileLoader, Constants, Promise, _, waterline, Log
             Constants.Profiles.Directory
         ).then(function (profiles) {
             var promises = _.map(profiles, function (contents, name) {
-                return self.put(name, contents);
+                return self.put(name, contents.toString());
             });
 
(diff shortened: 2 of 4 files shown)


benbp:
I deleted the hound errors (line too long, 'Buffer' not defined) because they are false negatives.

rolandpoulter:
:+1:

tldavies:
+1

lib/common/template.js
```diff
@@ -41,7 +41,7 @@ function templateServiceFactory(FileLoader, Constants, Promise, _, waterline, Lo
             Constants.Templates.Directory
         ).then(function (templates) {
             var promises = _.map(templates, function (contents, name) {
-                return self.put(name, contents);
+                return self.put(name, contents.toString());
```
rolandpoulter: Only concern here is that this will break if context is null or undefined. I'm assuming that is not possible here.

benbp: Yeah I think worst case we'll get an empty buffer out of the fs library.

benbp: Yeah, the fs library will give us an empty buffer for an empty file. If the file does not exist, we'll have failed before we get to this point:

```
> e = fs.readFileSync('./nothere')
Error: ENOENT, no such file or directory './nothere'
    at Object.fs.openSync (fs.js:432:18)
    at Object.fs.readFileSync (fs.js:289:15)
    at repl:1:9
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
    at Interface._line (readline.js:531:8)
    at Interface._ttyWrite (readline.js:760:14)
    at ReadStream.onkeypress (readline.js:99:10)
> e = fs.readFileSync('./emptyfile')
<Buffer >
> e.toString()
''
```

kacole2:
thanks. i can test it here soon

PR #49, opened 2015-12-18
Adds node ID caching layer to lookup service.

rolandpoulter:
https://groups.google.com/forum/#!topic/rackhd/CiRQhVdkE38

Attempting to improve logging performance by introducing caching to lookup-service. First cache miss results in a DB query, subsequent cache lookups for the same IP or MAC address will wait for the first query to respond unless the there is already a valid entry in the cache.

This should greatly reduce number DB queries by the LookupService.

### DIFF
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -71,7 +71,7 @@ function lookupServiceFactory(
      * @constructor
      */
     function LookupService () {
-      this.resetNodeIdCache();
+        this.resetNodeIdCache();
     }
 
     /**
@@ -96,19 +96,19 @@ function lookupServiceFactory(
                               otherwise null.
      */
     LookupService.prototype.checkNodeIdCache = function (key) {
-      if (typeof this.nodeIdCache[key] === 'string') {
-          return Promise.resolve(this.nodeIdCache[key]);
-      }
+        if (typeof this.nodeIdCache[key] === 'string') {
+            return Promise.resolve(this.nodeIdCache[key]);
+        }
 
-      if (Array.isArray(this.nodeIdCache[key])) {
-          return new Promise(function (resolve) {
-              this.nodeIdCache[key].push(resolve);
-          }.bind(this));
-      }
+        if (Array.isArray(this.nodeIdCache[key])) {
+            return new Promise(function (resolve) {
+                this.nodeIdCache[key].push(resolve);
+            }.bind(this));
+        }
 
-      this.nodeIdCache[key] = [];
+        this.nodeIdCache[key] = [];
 
-      return null;
+        return null;
     };
 
     /**
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -101,14 +101,15 @@ function lookupServiceFactory(
       }
 
       if (Array.isArray(this.nodeIdCache[key])) {
-          return new Promise(function (resolve, reject) {
+          return new Promise(function (resolve) {
               this.nodeIdCache[key].push(resolve);
           }.bind(this));
       }
 
       this.nodeIdCache[key] = [];
+
       return null;
-    }
+    };
 
     /**
      * Cache Node ID by either an IP or MAC address.
@@ -125,7 +126,9 @@ function lookupServiceFactory(
             });
         }
 
-        return this.nodeIdCache[key] = value;
+        this.nodeIdCache[key] = value;
+
+        return value;
     };
 
     /**
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -71,7 +71,62 @@ function lookupServiceFactory(
      * @constructor
      */
     function LookupService () {
+      this.resetNodeIdCache();
     }
+
+    /**
+     * Reset all cached Node IDs.
+     */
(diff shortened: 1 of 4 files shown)


spec/lib/services/lookup-spec.js
```diff
@@ -34,7 +34,54 @@ describe('Lookup Service', function () {
...
+          lookupService.checkNodeIdCache('testAddress').then(spy1, spy1);
+          lookupService.checkNodeIdCache('testAddress').then(spy2, spy2);
+        });
+
+        it('should resolve pending cache checks once a value is assigned', function () {
+          lookupService.assignNodeIdCache('testAddress', 'nodeId');
+          setTimeout(function () {
+            expect(spy1.called).to.be.ok;
+            expect(spy2.called).to.be.ok;
+            done();
```
benbp: Not passed in as an argument to the 'it' callback.

lib/services/lookup.js
```diff
@@ -71,9 +71,64 @@ function lookupServiceFactory(
...
+    };
+
+    /**
+     * Check the cache for a Node ID by IP or MAC addres.
+     * @param  {String} cache key - IP or MAC address.
+     * @return {Promise|null} Returns a promise if the Node ID is cached,
+                              otherwise null.
+     */
+    LookupService.prototype.checkNodeIdCache = function (key) {
+      if (typeof this.nodeIdCache[key] === 'string') {
```
benbp: Double spacing here.

spec/lib/services/lookup-spec.js
rolandpoulter: Already fixed this. Tests should pass now.

rolandpoulter:
Uggh. -0.02% code coverage loss.

benbp:
Two questions:

1. Is this missing cache refresh/expiration right now?
2. I assume that independent calls to a cache promise object from different places within the code will chain onto the shared reference to the object, yes? That way we avoid the promise multicast problem?

DavidjohnBlodgett:
I think code coverage will always go down if you delete anything~ even if its replaced with updated code~

rolandpoulter:
1) Correct, it doesn't try to populate the cache, nor does it every clear the cache. It should be trivial memory usage even with 1000s of nodes, but if it is a problem we can easily address it then.

2) I think the answer is yes. I'll double check with you.

heckj:
@rolandpoulter any reason not to use a the `lrucache` library with some options for timeout and decay of the cache over time - seems like that would have a smaller overall memory footprint, where this could conceivably grow memory usage without bound under pathological testing conditions (which we've experienced before)

rolandpoulter:
@heckj Nope, I didn't know about it. It should be pretty simple to change it to use `lru-cache` instead of a plain Object.

I'll update my PR soon.

rolandpoulter:
@benbp @heckj ^^

DavidjohnBlodgett:
:+1:

heckj:
:+1:

yyscamper:
@rolandpoulter :

The caching is good, I just wondering doesn't the mongodb have such caching capability? if so, in Mongodb or Waterline, we can have a common cache layer for all query.

And could you explain how you sync your cache with real DB data?

heckj:
@yyscamper MongoDB (like any database system) includes some caching, this is intended to cache *before* we make a request to mongoDB so that we don't need to incur the cost of that call every time we're using the lookup service.

LRU is specifically set to expire out data, and this code is set to attempt the cache first, and load on a cache miss, with a limited set of values. there's no explicit sync otherwise, so this does introduce the possibility that the DB can be updated and the cache will return a previous value that's not strictly correct. In the lookup service, the strict atomicity isn't necessary and the eventual consistency is quite acceptable.

lib/services/lookup.js
```diff
@@ -71,9 +73,76 @@ function lookupServiceFactory(
...
+     * Reset all cached Node IDs.
+     */
+    LookupService.prototype.resetNodeIdCache = function () {
+        if (this.nodeIdCache) {
+          this.nodeIdCache.reset();
+          return;
+        }
+
+        this.nodeIdCache = LRU({
+            max: 500
```
heckj: @benbp should we also be setting a `maxAge` key here in using the cache with the lookup services to limit how stale a cache value can be?

lib/services/lookup.js
```diff
@@ -71,9 +73,76 @@ function lookupServiceFactory(
...
+        this.resetNodeIdCache();
     }
 
     /**
+     * Reset all cached Node IDs.
+     */
+    LookupService.prototype.resetNodeIdCache = function () {
+        if (this.nodeIdCache) {
+          this.nodeIdCache.reset();
+          return;
```
anhou: indentation is 2, fixed to 4

rolandpoulter:
Fixed whitespace.

rolandpoulter:
@heckj @benbp I think this is ready to be merged.

heckj:
:+1: from me

rolandpoulter:
@benbp `lru-cache` was not being used before, so upgrading the version should not cause anything to break. Just double checked that.

benbp:
:+1:

PR #50, opened 2015-12-21
Align with the IEEE 802 format

scott-xu:
Align with the standard (IEEE 802) format for MAC-48 addresses
https://en.wikipedia.org/wiki/MAC_address#Notational_conventions


lib/serializables/mac-address.js
```diff
@@ -25,7 +25,7 @@ function MacAddressFactory (assert, Serializable) {
         properties: {
             value: {
                 type: 'string',
-                pattern: '^([0-9a-fA-F][0-9a-fA-F]:){5}([0-9a-fA-F][0-9a-fA-F])$'
+                pattern: '^([0-9a-fA-F]{2}[:-]){5}([0-9a-fA-F]{2})$|^([0-9a-fA-F]{4}[.]){2}([0-9a-fA-F]{4})$'
```
scott-xu: Why not buy a wider monitor

spec/lib/serializables/mac-address-spec.js
```diff
@@ -37,8 +37,16 @@ describe('MacAddress', function () {
 
     describe('validation', function () {
         describe('resolved', function () {
-            it('should resolve if valid', function () {
-                return this.subject.validate();
+            it('should resolve if it is six groups of two hexadecimal digits, separated by colons', function () {
```
scott-xu: Why not buy a wider monitor

spec/lib/serializables/mac-address-spec.js
```diff
@@ -37,8 +37,16 @@ describe('MacAddress', function () {
...
+                return new MacAddress({ value: '00:11:22:33:44:55' }).validate();
+            });
+            
+            it('should resolve if it is six groups of two hexadecimal digits, separated by hyphens', function () {
+                return new MacAddress({ value: '00-11-22-33-44-55' }).validate();
+            });
+            
+            it('should resolve if it is three groups of four hexadecimal digits separated by dots', function () {
```
scott-xu: Why not buy a wider monitor

heckj:
@scott-xu really? On every line? :-)

Seriously, though - the .jshintrc is our agreed upon styling format. THe line length is reasonably arbitrary, but not entirely - to support/make it as effective to develop with constrained resources as not (i.e. from within a terminal window)

scott-xu:
@houndci Sorry, robot. Am I another robot?

heckj:
@scott-xu a pretty good robot if you're changing code in a semantically consistent way

spec/lib/serializables/mac-address-spec.js
```diff
@@ -37,8 +37,16 @@ describe('MacAddress', function () {
 
     describe('validation', function () {
         describe('resolved', function () {
-            it('should resolve if valid', function () {
-                return this.subject.validate();
+            it('should resolve if it is six groups of two hexadecimal digits, separated by colons', function () {
+                return new MacAddress({ value: '00:11:22:33:44:55' }).validate();
+            });
+            
+            it('should resolve if it is six groups of two hexadecimal digits, separated by hyphens', function () {
```
scott-xu: Why not buy a wider monitor

heckj:
@scott-xu Code looks good, and no qualms with supporting the 802 spec, but there may be some fallout if you utilize that newer format (xxxx.yyyy.zzzz) simply because I'm not sure we were entirely consistent on how we were storing and using Mac-addresses. Also be aware that mongodb has some serious heartburn when attempting to store keys it's datastore with a "." character, so do plan on some deeper testing to fully enable that format Mac-address to make sure it's working

heckj:
(+1)

scott-xu:
A Mac address itself is a fixed value, however it can display in different formats. So I will store 112233445566 in database, but display it in different format for different scenario .

heckj:
@scott-xu works for me!

anhou:
@scott-xu  Do you have noticed another macaddress check in on-core library?  at https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L181 it could be updated at the same time

lib/common/constants.js
```diff
@@ -176,7 +176,7 @@ function constantsFactory (path) {
         Regex: {
             Base64: /^(?:[A-Z0-9+\/]{4})*(?:[A-Z0-9+\/]{2}==|[A-Z0-9+\/]{3}=|[A-Z0-9+\/]{4})$/i,
             Encrypted: /^(?:[A-Z0-9+\/]{4})*(?:[A-Z0-9+\/]{2}==|[A-Z0-9+\/]{3}=|[A-Z0-9+\/]{4})\.(?:[A-Z0-9+\/]{4})*(?:[A-Z0-9+\/]{2}==|[A-Z0-9+\/]{3}=|[A-Z0-9+\/]{4})$/i, // jshint ignore:line
-            MacAddress: /^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$/,
+            MacAddress: /^([0-9a-fA-F]{2}[:-]){5}([0-9a-fA-F]{2})$|^([0-9a-fA-F]{4}[.]){2}([0-9a-fA-F]{4})$/,
```
scott-xu: Why not buy a wider monitor

yyscamper: @scott-xu:
Many people like to use vertical screen to program, meanwhile, we need to view our code in many devices, even in some small screen devices, like iPad or Phone, of course, much more other reason.

Current, we limit the line character size to 100. We may can extend to 128, and don't want to be more high. @RackHD/corecommitters

yyscamper: Please add the "// jshint ignore:line" at the end of your regex expression, as it is not good to split the regex, but we can tell the jshint to explicitly ignore such warnning.

scott-xu: My iPad works pretty well even in portrait mode. Let me try iPhone later.

scott-xu: On my iPhone, I need to slide left/right. But it still works. If I choose desktop mode, I don't need to slide left/right. Instead, I can zoom in/out.

scott-xu: My suggestion is to set the max length to the same as github code page width so that
1. No horizontal scroll bar
2. No space waste

heckj:
-1: gates aren't passing: @scott-xu please fix the unit tests so they pass. See https://travis-ci.org/RackHD/on-core/jobs/98287278 for the explicit error. 

Also please take up a request to change our stylistic settings (line length) in email with the core team at rackhd@googlegroups.com, although the pictures in the PR were amusing.

scott-xu:
UT fixed.
I'm not a fan of stylist. I think that's all I can do for this PR.

jlongever:
Can one of the admins verify this patch?

benbp:
I'm going to close this for now. While it would be nice to conform to the IEEE format, there's not really any pressing need for it, and there's also potential risk that hasn't been investigated in regards to string formatting logic across the codebase and whatnot. Feel free to re-open if you want to fix the style violations and do some light testing of the system as a whole.

PR #51, opened 2015-12-22
log level reduction

heckj:
final pull request to resolve https://github.com/RackHD/RackHD/issues/34

**NOTE**: Requires all other PR's for log normalization in projects
using on-core to be merged prior to merging this PR. As I'm setting this
up, only https://github.com/RackHD/on-tasks/pull/95 remains to be merged
prior to this PR

### DIFF
--- a/lib/common/constants.js
+++ b/lib/common/constants.js
@@ -22,28 +22,18 @@ function constantsFactory (path) {
         },
         Logging: {
             Colors: {
-              emerg: 'red',
-              alert: 'yellow',
-              crit: 'red',
               critcal: 'red',
               error: 'red',
-              warning: 'red',
-              notice: 'yellow',
+              warning: 'yellow',
               info: 'green',
               debug: 'blue',
-              silly: 'blue'
             },
             Levels: {
-                emerg: 8,
-                alert: 7,
-                crit: 6,
-                critical: 6,
-                error: 5,
-                warning: 4,
-                notice: 3,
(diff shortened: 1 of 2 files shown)


yyscamper:
@heckj : I see some "logger.deprecate" in code, now it's a special function compared with other log levels, Do you think we'd better move the deprecate to the levels table?

If add into levels table, seems the deprecate is different with "error, warn, ...", hard to determine its priority.
But the benefit is we can get uniform control over all logger behavior.

heckj:
@yyscamper I think keeping `deprecate` as a special function is valuable and has merit as it is, rather than trying to shove it into a level list, that has semantically different meaning. `deprecate` specifically maps to a log level as well, so I view it as a useful convenience function that should be primarily for ourselves. I don't believe it violates any logging behaviour given that it's mapped to a specific log level and utilizes all the same underlying infrastructure

yyscamper:
@heckj , I agree. :+1:

anhou:
the enum number is '1.2, 4, 5 ,6', could you help keep it in sequential? to avoid when someone see these numbers, have the question "why skip 0 and 3"
:+1:  after modify to sequential numbers

heckj:
@anhou updated to sequence the values linearly and remove the "gaps" from a straight up delete

anhou:
:+1: 
https://github.com/RackHD/on-tasks/pull/95 has been merged, so I'll merge this PR

PR #52, opened 2015-12-24
Issue#47: fix lookups doesn't return all records for empty query

yyscamper:
Fix [Issue#47](https://github.com/RackHD/RackHD/issues/47).

Any detail, please go to that issue page.

@RackHD/corecommitters @iceiilin @WangWinson @cgx027 @pengz1 @sunnyqianzhang

### DIFF
--- a/lib/models/lookup.js
+++ b/lib/models/lookup.js
@@ -34,13 +34,17 @@ function LookupModelFactory (Model, assert, Errors, Promise, Constants) {
             }
         },
         findByTerm: function (term) {
-            return this.find({
-                or: [
-                    { node: term },
-                    { macAddress: term },
-                    { ipAddress: term }
-                ]
-            });
+            var query = {}; //empty query will find all entries
+            if (term) {
+                query = {
+                    or: [
+                        { node: term },
+                        { macAddress: term },
+                        { ipAddress: term }
+                    ]
+                };
+            }
+            return this.find(query);
         },
         findOneByTerm: function (term) {
             return this.findByTerm(term).then(function (records) {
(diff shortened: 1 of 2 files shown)


iceiilin:
+1

lib/models/lookup.js
```diff
@@ -34,13 +34,17 @@ function LookupModelFactory (Model, assert, Errors, Promise, Constants) {
...
         },
         findByTerm: function (term) {
-            return this.find({
-                or: [
-                    { node: term },
-                    { macAddress: term },
-                    { ipAddress: term }
-                ]
-            });
+            var query = {}; //empty query will find all entries
```
WangWinson: if ```term``` is empty, ```findByTerm``` will find nothing sounds reasonable. 
The better place to add this logic maybe in on-http/lib/api/lookups.js. 
When ```req.query.q``` is empty, call ```find``` directly.

heckj:
:+1: Felix, perfect - thank you!

lib/models/lookup.js
anhou: @WangWinson it seems  when input is emtpy, waterline return all is the default behavior, so it's reasonable here

anhou:
:+1:

lib/models/lookup.js
yyscamper: I prefer to keep the behavior same with waterline, for many built-in function of waterline, it will tend to handle all records if the query is empty.
In on-http, sometimes we use the waterline built-in function, sometimes use the function that we customized for each data model,  so a consistent behavior will be better.

PR #53, opened 2015-12-24
Waterline updates

heckj:
While digging around looking at how we're reconnecting automatically on MongoDB, I saw we were downlevel and had some dead code enabled, so clipping things out and upgrading us to sails-mongo 11.5, which resolves https://github.com/RackHD/RackHD/issues/56.

Notes from @frey indicate that later versions of sails-mongo may not properly support the native drivers on unix, so taking a stab at this while I'm digging out the specifics of what is and isn't supported there.

According to https://docs.mongodb.org/ecosystem/drivers/node-js/, the underlying mongodb driver of 2.0.42 supports our versions of MongoDB and of Node.js:

- https://docs.mongodb.org/ecosystem/drivers/driver-compatibility-reference/#reference-compatibility-language-node
- https://docs.mongodb.org/ecosystem/drivers/driver-compatibility-reference/#reference-compatibility-mongodb-node

The only thing the docs assert it doesn't include is support for MongoDB 3.2

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ module.exports = function (di, directory) {
                 helper.requireWrapper('waterline', 'Waterline'),
                 helper.requireWrapper('waterline-criteria', 'WaterlineCriteria'),
                 helper.requireWrapper('sails-mongo', 'MongoAdapter'),
-                helper.requireWrapper('sails-disk', 'DiskAdapter'),
+                // helper.requireWrapper('sails-disk', 'DiskAdapter'),
                 helper.requireWrapper('amqp', 'amqp'),
                 helper.requireWrapper('domain', 'domain'),
                 helper.requireWrapper('node-uuid', 'uuid'),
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -21,7 +21,7 @@ module.exports = function (di, directory) {
                 helper.requireWrapper('waterline', 'Waterline'),
                 helper.requireWrapper('waterline-criteria', 'WaterlineCriteria'),
                 helper.requireWrapper('sails-mongo', 'MongoAdapter'),
-                helper.requireWrapper('sails-disk', 'DiskAdapter'),
+                // helper.requireWrapper('sails-disk', 'DiskAdapter'),
```
anhou: Don't just comment out, remove it
:+1:  after removing

heckj:
@anhou thanks, missed that one - removed the commented out code and rebased the whole stack on latest master

anhou:
+1

benbp:
Yep, you were right, DiskAdapter is no longer used (https://github.com/RackHD/on-core/commit/cca1c1f3827509bcc45c88f90f2be2a364082ef9).

benbp:
:+1:

PR #55, opened 2015-12-29
add node relation mapping table

iceiilin:
Add node relation mapping table.
I listed relation and its class ("link" and "component") here. It is used currently only for deleting node and updating its target nodes, but might extent to other actions (CRUD) in the future.

We can classify node relations into two classes: "component" and "link". If the relation is "component", its subordinate nodes will be deleted (like enclosure and compute); if the relation is "link", the target nodes will only be updated (like pdu and compute).

related PR:
https://github.com/RackHD/on-http/pull/80

### DIFF
--- a/lib/common/constants.js
+++ b/lib/common/constants.js
@@ -201,48 +201,56 @@ function constantsFactory (path) {
             ]
         },
         NodeRelations: {
+            /* There are two classes about the pair of relations:
+             * component: one node is a component of the other
+             *     e.g.  compute node vs. enclosure node
+             *     expect behavior:
+             *         the "By" in the relationType is used to tell component from its parent
+             *         delete a node as well as its components;
+             *         delete a component and update its parenet node
+             * link: two nodes is linked to each other
+             *     e.g.  compute node vs pdu node (TODO)
+             *     expect behavior:
+             *         delete one node and update the other
+            */
             encloses: {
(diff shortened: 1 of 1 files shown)


heckj:
@iceiilin Is this meant to be a list of all relationships we use, or the relationships on which we delete related nodes?

For example, based on your comment and the fact that "powered by" is in here, I'd presume that if we deleted a PDU, that all nodes powered-by that PDU would likewise be deleted. Is that the intent, or is this meant to be a catalog of allowed relationships?

Where is definition that the business logic of related elements getting deletion? Is it intended to be defined in this constant, or is it in business logic in on-http, and which just references this map as a list of possible relationships?

iceiilin:
@heckj I listed relation and its class ("link" and "component") here. It is used currently only for deleting node and updating its target nodes, but might extent to other actions (CRUD) in the future.

"relationClass" in this constant defines the business logic of deletion, as I described in the comments in constants.js. After internal discussion, we can classify node relations into two classes: "component" and "link". If the relation is "component", its subordinate nodes will be deleted (like enclosure and compute); if the relation is "link", the target nodes will only be updated (like pdu and compute).

The on-http get this relationClass from on-core constants, and delete/update nodes accordingly.

yyscamper:
:+1:

anhou:
:+1:

PR #61, opened 2016-01-05
Upgrade lodash to version 3.10.1

rolandpoulter:
This should be merged before the `on-http`, `on-syslog`, `on-tasks`, and `on-taskgraph` upgrades are merged. Those PRs will be opened shortly after this.

According to lodash changelog the `.flatten()` was changed to be shallow, so `.flattenDeep()` is now used instead to prevent breakage. Also when wrapping an object with lodash and then calling `.forEach()` will not iterate until `.value()` is called. It was changed to be lazily evaluated.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,7 @@ module.exports = function (di, directory) {
 
     return {
         helper: helper,
-        injectables: _.flatten(
+        injectables: _.flattenDeep(
             [
                 // NPM Packages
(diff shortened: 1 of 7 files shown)


benbp:
:+1:

yyscamper:
:+1: Since the 1.1.0 branch is just created, so this is a safe time window to do such lodash upgrade.

PR #63, opened 2016-01-12
Add unlink method to TemplateService

zyoung51:


lib/common/template.js
```diff
@@ -108,5 +108,18 @@ function templateServiceFactory(FileLoader, Constants, Promise, _, waterline, Lo
         });
     };
 
+    TemplateService.prototype.unlink = function (name, scope) {
```
heckj: why `unlink` and not `destroy` or `remove`?

heckj:
question about naming, but otherwise :+1:

heckj:
er, could also use unit tests :-)

lib/common/template.js
```diff
@@ -108,5 +108,18 @@ function templateServiceFactory(FileLoader, Constants, Promise, _, waterline, Lo
...
+        scope = scope || 'global';
+        return new Promise(function(resolve, reject) {
```
benbp: You can just do:

```
return waterline.templates.destroy(...);
```

Rather than wrapping it in a Promise constructor, since all the waterline model methods are already promisified.

lib/common/template.js
benbp: unlink is consistent with the node `fs` module's naming convention, so perhaps that's why?

lib/common/template.js
zyoung51: FWIW, that didn't work, which is why it is wrapped this way.  The promise would resolve and the database entry would never be removed.

benbp: Interesting, could be a bug or some weird behavior with waterline, let me try to repro.

lib/common/template.js
zyoung51: yep, that's why.  The unlink here is part of the unlink of the actual file during sku pack deletion.

lib/common/template.js
benbp: It also could be related to the bluebird/not bluebird promise thing, but I thought we had resolved that (maybe not?).

benbp: Chatted on slack, appeared related to the fact that Bluebird doesn't coerce an array of promises by default without prompting via `.map`, `.all` or `.spread`.

zyoung51: Will update with the methods we talked about and then reopen a new PR.

PR #65, opened 2016-01-13
Added script and secure variable for dependent builds

tldavies:


tldavies:
@jlongever - Here is the PR for when on-core merges to master it reruns all the depends

build_children.sh
```diff
@@ -0,0 +1,30 @@
...
+curl -X POST https://api.travis-ci.org/builds/$BUILD_NUM_SYSLOG/restart --header "Authorization: token "$AUTH_TOKEN
+
+#Echo out what builds where restarted
+echo Restarted on-dhcp-proxy Build = $BUILD_NUM_DHCP
+echo Restarted on-tasks = $BUILD_NUM_TASKS
+echo Restarted on-tftp Build = $BUILD_NUM_TFTP
+echo Restarted on-syslog Build = $BUILD_NUM_SYSLOG
+
+
+echo "Note - on-core restarts builds for on-dhcp-proxy, on-tftp, on-syslog, and on-tasks. When on-tasks succeeds, it restarts builds for on-taskgraph and on-http. "
```
tldavies: @benbp Does this work as a note?

build_children.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+#Gets the latest Master build
+GetMasterBuild () {
+
+    BUILD=$(curl -s 'https://api.travis-ci.org/repos/RackHD/'$1'/branches/master' | grep -o '{"branch":{"id":[0-9]*,' | grep -o '[0-9]' | tr -d '\n')
```
benbp: I don't think you need or want quotes around `$1` here.

build_children.sh
benbp: Yes, thanks.

benbp:
Do we end up running all the downstream builds 2x because of the node 4 and node 10 builds?

jlongever:
Ignore that jenkins failure for now.

tldavies:
@benbp Yes it tries to run them all twice cause of Node 4 and Node 10 builds. If it is already running the build though you get a fail error from the api. 

Error is build is already running: {"result":false,"flash":[{"error":"This build currently can not be restarted."}]}

tldavies:
@RackHD/corecommitters

build_children.sh
tldavies: It didn;t work when there wasn't quotes.

benbp:
@tldavies so if the node 4 build runs first, then we won't actually succeed in running the node 10 build at all?

benbp:
Is there enough metadata from the API to grab both the most recent node 10 and node 4 builds and restart each individually?

jlongever:
Question.. how does the on-core PR commit get applied to the downstream build?  I see the travis API calls to restart the last build for the dependencies but it's not clear how the on-core commit is applied to those builds.

tldavies:
@benbp The Travis build runs both Node 4 and 10. So when it is restarted it starts both again. Node 4 vs Node 10 are different jobs but the same build. If we only ran the job than it would only run Node 10 or Node 4.

tldavies:
@jlongever So this will only work when something is merged to on-core master. As forked PRs cannot access the environment variables. And then the dependent builds pull on-core from github which will have the newest version.

jlongever:
I think the goal should be to prevent changes from being merged in that potentially break dependencies.

benbp:
@jlongever yeah, it kind of sucks to not be able to do dependent PR builds before merge, but I think the tradeoff might be better than all the false negatives we've been getting from Jenkins at the same time.

jlongever:
not really sure what the value is other than builds happening faster?   the consistency issues with jenkins are solvable.

benbp:
If they're solvable, then yeah, we should probably just plan on migrating to Jenkins fully.

benbp:
I do think builds happening fast has a high value though as an organization's commit velocity increases.

benbp:
:+1:

jlongever:
test this please

VulpesArtificem:
:+1:

tldavies:
test this please

jlongever:
test this please

tldavies:
test this please

jlongever:
test this please

PR #66, opened 2016-01-19
improve log output (color/level/caller)

yyscamper:
Change includes:

- default turn off colorful output, user can turn on the colorful output by configuration: --color or --logColorEnable
- create argument handler service as the common place for argument handling. 
- print full log level name
- print caller uses the same color of corresponding log level
- change the color for 'debug' level to cyan
- improve LogEvent unittest coverage

this change is from the google group discussion [Remove the color code from RackHD log file]( https://groups.google.com/forum/#!topic/rackhd/7xtNfbg9jzM)
The code follows the implementation of the "colors" module: https://github.com/Marak/colors.js/blob/master/lib%2Fsystem%2Fsupports-colors.js#L34-L38

@RackHD/corecommitters @iceiilin @WangWinson @cgx027 @pengz1 @sunnyqianzhang

### DIFF
--- a/lib/common/constants.js
+++ b/lib/common/constants.js
@@ -26,7 +26,7 @@ function constantsFactory (path) {
               error: 'red',
               warning: 'yellow',
               info: 'green',
-              debug: 'blue',
+              debug: 'cyan',
             },
             Levels: {
                 critical: 5,
--- a/lib/serializables/log-event.js
+++ b/lib/serializables/log-event.js
@@ -32,7 +32,10 @@ function LogEventFactory (
     pretty,
     console
 ) {
-    colors.setTheme(Constants.Logging.Colors);
+    //default turn off the colorful output to achieve tidy and clean log file.
+    LogEvent.colorEnable = false;
+
+    colors.setTheme(Constants.Logging.Colors); //setTheme will not impact color enable or not
 
     function LogEvent(defaults) {
         Serializable.call(this, LogEvent.schema, defaults);
@@ -80,9 +83,8 @@ function LogEventFactory (
         var statement = [];
 
         this.context = LogEvent.redact(this.context);
-
-        statement.push(this.level[0].toUpperCase());
-        statement.push(this.timestamp);
+        statement.push('[%s]'.format(this.level));
+        statement.push('[%s]'.format(this.timestamp));
         statement.push('[%s]'.format(this.name));
         statement.push('[%s]'.format(this.module));
         statement.push('[%s]'.format(
@@ -93,13 +95,16 @@ function LogEventFactory (
         var output = statement.join(' ');
 
         console.log(colors[this.level](output));
-        console.log(' -> ' + this.caller);
+        console.log(colors[this.level](' -> ' + this.caller));
 
         if (_.keys(this.context).length > 0) {
             console.log(
                 pretty.render(
                     this.context,
-                    { noColor: false }
+                    {
+                        noColor: !LogEvent.colorEnable,
+                        numberColor: 'cyan'
+                    }
                 )
             );
         }
@@ -117,6 +122,15 @@ function LogEventFactory (
             }).then(function (object) {
                 return object.validate();
             });
+    };
+
+    /**
+     * Set enabled or disabled for colorful log
+     * @param {boolean} enable - true to enable colorful log; false to disable
+     */
+    LogEvent.setColorEnable = function(enable) {
+        LogEvent.colorEnable = (enable ? true : false);
+        colors.enabled = LogEvent.colorEnable;
     };
 
     LogEvent.sanitize = function(context) {
--- a/lib/services/argument-handler.js
+++ b/lib/services/argument-handler.js
@@ -1,0 +1,60 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = argumentHandlerServiceFactory;
+
+argumentHandlerServiceFactory.$provide = 'Services.ArgumentHandler';
+argumentHandlerServiceFactory.$inject = [
+    '_',
+    'Services.Configuration',
+    'LogEvent'
+];
+
+function argumentHandlerServiceFactory(
+    _,
+    config,
+    LogEvent
+){
+    function ArgumentHandlerService() {}
+
+    /**
+     * Get the value from configuration by a collection of keys
+     *
+     * Sometimes we define different keys in configuration file and command line argument, because
+     * configuration file prefers the full key name, but command line prefers the short and mnemonic
+     * name. The argument handle can support the configuration from either file or command line, so
+     * this function is a helper to enumerate all keys, any key exists in configuration will return
+     * its value, if all keys don't exist, then return the default value.
+     * The earlier key in the input keys collection will take precedence over latter.
+     *
+     * @param {Collection} keys - the collection of keys
+     * @param {*} defaultValue - the default value that will be returned if all keys don't exist.
+     * @returns {*} the value of those keys, if all keys don't exist, will return the defaultValue.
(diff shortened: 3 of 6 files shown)


yyscamper:
The travis ci should have a false alarm, I restarted the build, then it success.

lib/serializables/log-event.js
```diff
@@ -32,7 +32,18 @@ function LogEventFactory (
     pretty,
     console
 ) {
+    //default turn off the colorful output to achieve tidy and clean log file.
+    //however user can still enable it via command line argument.
+    var colorEnable = false;
+    var argv = process.argv;
```
heckj: It seems very unintuitive (and likely to be lost in the future) to know that we're inspecting arguments to the nodes process within this module. While this would function, command line options would be much better suited to being captured in a single location - I recommend in the top-level index.js - and relevant options then set internally or passed into this module to enable/disable color.

yyscamper: @heckj : Are you suggesting designing a function (such as **setColorEnable**) in LogEvent. Then in each app's index.js (on-http/on-taskgraph/on-dhcp-proxy/on-syslog/on-tftp), call this function to set color mode bases on command line options)?
 
I agree bury those command line options in the LogEvent is not a good idea. But to avoid cross repositories change and multi code copying, how about define a common argv processor in on-core then execute the common processor in on-core index.js or call the common argv processor in each app index.js.

heckj: yeah, something like that would work. Start with a default of off in this module (like you have) and enable a method to toggle it on/off. That can then be called within index.js to enable/disable as needed, and would set up a good pattern if we want/need to add additional command line options.

Be aware that nconf(https://www.npmjs.com/package/nconf) in the configuration service is *also* inspecting the process, and will set configuration values based on what it finds in those parameters. We'll want to at least be aware of potential overlapping impact there: https://github.com/RackHD/on-core/blob/master/lib/services/configuration.js#L39

Yes, I agree a command-line mechanism common to all the applications would be appropriate, so something in on-core would work. In the past we've used the library https://github.com/tj/commander.js/ for this, which proved pretty useful. It would also mean changing up all of the application index.js to do the argument processing for each application first, then driving the relevant application functionality (which is usually starting a listening service)

yyscamper: I planed to use the configuration service to define the log color option, but stopped it because let the LogEvent depends on the Configuration Service is not good design, so that's my compromised design by only examining the process.env.

Since nfcon already process the command line options, so we don't need the "commander" though it is a very powerful library even only look at its star number :). Move to index.js will not have the dependency concern.

Our design of Configuration Service should be a combination of handling arguments from command line options, system environment and config file.

So I will leverage the Configuration Service to define the log color switch, so user can either enable  the log color from command line option or from monorail.json.

jlongever:
test this please

jlongever:
test this please

yyscamper:
@heckj : My new change move the argument handler to `on-core/index.js`, so it can be shared by all RackHD app.
I know `on-core/index.js` is not a good place for argument handle especially after we have lots of argument need be processed.
Actually, the configuration.start (https://github.com/RackHD/on-core/blob/master/lib/services/configuration.js#L96-L100) should be the best place,  but unfortunately
it introduce `cyclic dependency`.
```
/home/onrack/src/on-taskgraph/node_modules/di/dist/cjs/injector.js:166
      throw new Error(("Cannot instantiate cyclic dependency!" + resolvingMsg)
            ^
Error: Cannot instantiate cyclic dependency! (TaskGraph.Runner -> TaskGraph.Subscriptions -> Protocol.TaskGraphRunner -> Services.Messenger -> Messenger -> Services.StatsD -> Services.Configuration -> LogEvent -> Services.Lookup -> Services.Configuration)
```

My next plan is to mimic [commander](https://www.npmjs.com/package/commander) to implement a class ArgumentHandler, every app can define its own argument handler, and these argument handler can be nested together, just like:
```
var hcore = new ArgumentHandler();
hcore.optoins('-c').action(function() ...)

var happ = new ArgumentHanlder();
happ.use(hcore);
happ.options('-x').action(function() ...)
```
This still need lot of coding work, so I want to keep this compromised design right now so we can enable the functionality earlier, then I will open another PR in future to include my above proposal design.

yyscamper:
I have opened an issue for houndci for its strange warning: https://github.com/thoughtbot/hound/issues/1040

yyscamper:
The houndci problem has been fixed: https://github.com/thoughtbot/hound/issues/1040, but it still need another PR change to trigger the houndci build again.

tldavies:
test this please

benbp:
@yyscamper if you just rebase locally to change the commit hashes, then force push your same changes to the branch, it will re-trigger a travis build I believe.

yyscamper:
I push a new commit for better code architecture. I create a ArgumentHandler service and leave it as the common place to handle all input arguments.

lib/services/argument-handler.js
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = argumentHandlerServiceFactory;
```
jlongever: @yyscamper This looks like a new service.  Where does the ArgumentHandlerService get used?

lib/common/constants.js
```diff
@@ -26,7 +26,7 @@ function constantsFactory (path) {
               error: 'red',
               warning: 'yellow',
               info: 'green',
-              debug: 'blue',
+              debug: 'cyan',
```
heckj: @benbp totally stupid question - can you visually discern the differences here, or does this make it appear the same as the other colors?

benbp: not a stupid question, let me find out!

benbp:
@yyscamper yep the cyclic dependency thing is not good. there's a lot of general refactoring that I think on-core could benefit from in this regard.

lib/common/constants.js
benbp: Heh, so, I can tell the difference between cyan and the other colors, however I **did** notice that I actually can't tell the difference between info and warning messages (green and yellow)!

lib/services/argument-handler.js
yyscamper: in the index.js of on-core, so it can take effect for all app

Felix

-----Original Message-----

lib/common/constants.js
yyscamper: For my eye, I just see the green color in terminal with dark background is not so visible, but the color "cyan" is comfortable in both dark and light background, so that's why I change this.


Felix

-----Original Message-----

yyscamper:
Yes, some of our modules' dependancy is not good. Take LogEvent for example, I think this is a basic and very low level module, because every other module may use logger to print information, but it depends on service.lookup and others, the dependancy reverse? :)

I agree we need some code refraction to sort out our dependancy chain.

Felix

-----Original Message-----

iceiilin:
+1

lib/common/constants.js
yyscamper: @benbp @heckj : are you against such color change?

benbp: @yyscamper no not at all, this change is fine. I happen to be colorblind, so @heckj was just making sure I could actually tell the difference between cyan and green!

benbp:
@yyscamper yeah the LogEvent is a tricky case, because we do use lookups as a way of injecting node data into the log headers so that they can be filtered by things like AMQP, elasticsearch, etc. I think the only way to solve this is to figure out some different sort of dependency structure (I'd rather not do proxy objects though), or a way to do circular dependency detection/loading.

benbp:
:+1:

PR #67, opened 2016-01-20
fix houndci false alarm about undefined variable

yyscamper:
Fix the annoying Hound CI false alarm for undefined variables by adding those variables into the jshint "predef" tables. (previous jshint "globals" options doesn't work well for houndci, but I tested "predef" works well, see my test PR: https://github.com/yyscamper/on-tasks/pull/10)

However, by introducing this change, there will be a new false alarm emerges:
```
Redefinition of '_'
```
This is caused by you use lodash not via injectable table but directly by require('lodash').
This false alarm usually only happens in "index.js", so it is rare to happen.

@RackHD/corecommitters @WangWinson @iceiilin

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -57,21 +57,18 @@
 
     // Environments
     "node": true,
+    "mocha": true,
 
     //Custom Globals
-    "globals" : {
-        "describe": false,
-        "it": false,
-        "before": false,
-        "beforeEach": false,
-        "after": false,
-        "afterEach": false,
-        "expect": false,
-        "should": false,
-        "request": false,
-        "helper": false,
-        "chai": false,
-        "sinon": false,
-        "Map": false
-    }
+    "predef" : [
+        "_",
+        "chai",
+        "expect",
+        "helper",
+        "Promise",
+        "request",
+        "should",
+        "sinon",
+        "sinonPromise"
+    ]
 }


yyscamper:
Wondering why houndci report error parsing config for jshint only in on-core repo, the same config file in other repos don't have such error.

spec/lib/common/file-loader-spec.js
```diff
@@ -79,7 +79,7 @@ describe('FileLoader', function () {
                 fs.readFileAsync.resolves('getAll');
                 fs.readdirAsync.resolves(['/tmp/foo']);
                 fs.statAsync.resolves({ isDirectory: function() { return true; } });
-                return this.subject.getAll('/tmp').then(function (files) {
+                return this.subject.getAll('/tmp').then(function() {
```
anhou: why remove 'files"?

anhou:
@yyscamper please fix the houndci

spec/lib/common/file-loader-spec.js
yyscamper: The `files` is not referenced anywhere, there will be houndci warning.

yyscamper:
@anhou I tried to fix the houndci error "Error parsing config for: jshint" but failed, I don't know why the houndci reports some error as I uses the same `.jshintrc` and `.hound.ym` with other repos (expect on-http), but those repos doesn't have such error.

I guess this may be houndci false alarm, but cannot make sure.

yyscamper:
I enabled the houndci in my forked repo, I tested not change the .jshintrc, only copy an unitest file, it shows the houndci still complains "Error parsing config for: jshint", please checked this PR:
https://github.com/yyscamper/on-core/pull/2, so it proves it's not my change introduces the houndci complain.

I guess maybe it is the file format problem (dos file / linux file/...), so I find a tool compared the `.hound.yml` and `.jshintrc` with the ones in on-tasks repo in binary level per byte, the tool shows they are exactly the same file even per byte.

yyscamper:
I finally resolved it, I reverted my change to `spec/lib/common/file-loader-spec.js` and keep this PR only change the `.jshintrc`, now houndci becomes work well. Wow!!

houndci is a strange and naughty robot !

anhou:
@yyscamper Thanks, :+1:

iceiilin:
+1

PR #70, opened 2016-01-29
Use array for exit code in child process

iceiilin:
Change accepted exit code in child process from a number to an array, which is align with bootstrap.js running in the node.

### DIFF
--- a/lib/common/child-process.js
+++ b/lib/common/child-process.js
@@ -49,12 +49,13 @@ function ChildProcessFactory (
         var self = this;
 
         assert.string(command);
+        assert.optionalArrayOfNumber(code);
 
         self.command = command;
         self.file = self._parseCommandPath(self.command);
         self.args = args;
         self.environment = env || {};
-        self.exitCode = code || 0;
+        self.exitCode = code || [0];
         self.maxBuffer = maxBuffer || Constants.ChildProcess.MaxBuffer;
 
         if (!self.file) {
@@ -136,7 +137,7 @@ function ChildProcessFactory (
         self.spawnInstance = nodeChildProcess.execFile(
                 self.file, self.args, options, function (error, stdout, stderr) {
 
-            if (error && error.code !== self.exitCode) {
+            if (error && self.exitCode.indexOf(error.code) === -1) {
                 self.hasBeenKilled = true;
                 logger.error('Error Running ChildProcess.', {
                     file: self.file,
@@ -145,7 +146,6 @@ function ChildProcessFactory (
                     stderr: stderr,
                     error: error
                 });
-
                 self.reject(error);
             } else {
                 self.hasBeenKilled = true;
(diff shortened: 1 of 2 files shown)


iceiilin:
The coverage decrease is unclear and I can't find details about it. The child-process.js I modified in this PR has 100% line and statement coverage.

iceiilin:
I have modified exit code parameter within RackHD, and there is only one place. https://github.com/RackHD/on-tasks/pull/114/files#diff-1c4c0a36d5f79178ac5aaa894f289424R38
So I can't think of the reason of Jenkins failure. Is there any log that I can look into?

heckj:
Hi @iceiilin  - why change the result code to be an array? It's not clear to me how you would leverage or use more than one response, and this subtly changes the implicit "truthiness" of the results, which could result in very unexpected responses. 

-1 until it's clear what the value is of adding this complicated response

benbp:
@heckj from looking it appears we may have had a bug before this given that, as @iceiilin mentions, bootstrap.js treats acceptedResponseCodes as an array (https://github.com/RackHD/on-http/blob/master/data/templates/bootstrap.js#L142). Since it does an indexOf call, I could see some false positives, for example where the error code is '127' and the acceptedResponseCodes is '1'.

That said, perhaps it would be better to fix bootstrap.js to treat the accepted response value as a string/number instead of an array?

heckj:
Thanks Ben, that helps me understand. I'll withdraw my -1 if the core committers would prefer to have this return an array result to match the bootstrap.js code, but I agree that updating/fixing bootstrap.js would seem to make more sense from a longer term fix perspective. There could easily be some reason I'm missing on why changing bootstrap.js would be a bad thing though.

jlongever:
test this please

yyscamper:
@benbp:

I haven't understood your comment about false positive case could you please explain more?

    Since it does an indexOf call, I could see some false positives, for example where the error code is        '127' and the acceptedResponseCodes is '1'.

@benbp, @heckj : I'd prefer to use array, Normally the command line returns 0 means it executes successfully, but `acceptedResponseCodes` is designed to intentionally ignore some particular errors, since we may want to ignore multi kind of errors, so Array should be better.

benbp:
@yyscamper right now the bootstrap.js implementation is buggy if used against a string instead of an array, for example `'127'.indexOf('1')` returns a truthy value even though '127' != '1'.

yyscamper:
Thanks @benbp, understand your concern, this is sure a problem, force it to array of number should fix this.

iceiilin:
@benbp I would prefer to use an array of number as accepted exit codes, to cover cases where several exit codes is acceptable. https://github.com/iceiilin/on-core/blob/bugfix/odr-399-ipmi-cata/lib/common/child-process.js#L52 has ensured the input code is an array of number. So I think this PR is ready. I will open another PR in on-http to solve the problem in bootstrap.js.

heckj:
changing from previous -1 to :+1:

yyscamper:
:+1:

PR #74, opened 2016-01-31
Refactor on-core unit-tests

jlongever:
- Remaining protocols tests
- Remove rabbitmq-server dependency.

@RackHD/corecommitters @keedya @zyoung51 @uppalk1 @tannoa2

### DIFF
--- a/spec/lib/protocol/waterline-spec.js
+++ b/spec/lib/protocol/waterline-spec.js
@@ -4,26 +4,36 @@
 'use strict';
 
 describe('Protocol.Waterline', function() {
-    var waterlineProtocol;
-    var collection = { identity: 'dummy' };
-
+    var waterlineProtocol,
+        collection = { identity: 'dummy' },
+        testSubscription,
+        messenger;
+        
     helper.before();
 
     before(function () {
         waterlineProtocol = helper.injector.get('Protocol.Waterline');
+        messenger = helper.injector.get('Services.Messenger');
+        var Subscription = helper.injector.get('Subscription');
+        testSubscription = new Subscription({},{});
+        sinon.stub(testSubscription);
+        sinon.stub(messenger);
     });
 
     helper.after();
 
     it('should publish a created event', function() {
+        messenger.publish.resolves();
         return waterlineProtocol.publishRecord(collection, 'created', { id: 1 });
     });
 
     it('should publish an updated event', function() {
+        messenger.publish.resolves();
         return waterlineProtocol.publishRecord(collection, 'updated', { id: 1 });
     });
 
-    it('should publish a destroyed event', function() {
+    it('should publish a destroyed event', function() {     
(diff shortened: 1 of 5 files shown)


keedya:
:+1:

jlongever:
test this please

spec/lib/protocol/waterline-spec.js
```diff
@@ -4,26 +4,37 @@
...
+        var Subscription = helper.injector.get('Subscription');
+        testSubscription = new Subscription({},{});
+        sinon.stub(testSubscription);
+        sinon.stub(messenger);
     });
 
     helper.after();
 
     it('should publish a created event', function() {
+        messenger.publish.resolves(Promise.resolve());
```
yyscamper: Is  this and following line intentionally? I don't see any assertion statement.

benbp: Promise.resolve is redundant regardless. You can get away with `messenger.publish.resolves()`

jlongever: @yyscamper nothing to assert?  This is the original code, only adding the publish() stub. 
@benbp just added it for the sake of clarity.

yyscamper: yes, no assertion here. @jlongever , I just see you don't touch those code, if could, it is a good chance to add some assertion statement here. Of course, add it to TODO list and do it in another PR is OK for me.

yyscamper: I just think if we don't do it this time, maybe no one will raise this up again

yyscamper:
@jlongever , those houndci warning is positive, you'd better fix them.

spec/lib/protocol/waterline-spec.js
jlongever: @yyscamper I meant what do you want to assert here?

benbp: @jlongever probably something like: `return expect(waterlineProtocol.publishRecord(collection, 'updated', { id: 1 })).to.be.fulfilled`

benbp: Functionally it's the same since a promise rejection will fail the test, but this way the code makes it a little clearer what the intention of the test is.

jlongever:
@benbp thanks.   Is this a show stopper?

benbp:
@jlongever not on the expect statement thing, but the resolves() thing would be nice to fix, because it's actually returning a new promise object as the value to the promise. Right now it's of no consequence because we aren't testing the messenger.publish resolution value, but I could see it potentially leading to test failures down the road. If you don't mind just doing a search and replace over your changes?

Otherwise +1

jlongever:
@benbp thanks.  updated.

benbp:
:+1:

jlongever:
test this please

PR #75, opened 2016-02-03
Workflow engine refactor

benbp:
Refactoring the workflow engine for:

- High Availability
- Fault Tolerance
- Better performance
- Better execution strategy

This PR primarily includes adding a mongo based store and an AMQP based messenger plugin layer for the workflow engine, and also moves/adds all the TaskGraph creation and validation code into this repository, so that it can be called from on-http where we do up front validation on request now.

Relies on RackHD/on-taskgraph#56, https://github.com/RackHD/on-tasks/pull/116, and RackHD/on-http#112

See the description of RackHD/on-taskgraph#56 for architectural slides.


benbp:
Mentioning a mix of those who came to the presentation/should review this: @RackHD/corecommitters @VulpesArtificem @heckj @king-jam @johren @stuart-stanley @jimturnquist @larry-dean @davequick @zyoung51 @zhangy42 @mattgam @tldavies @DavidjohnBlodgett @rolandpoulter

benbp:
Looks like coverage went down quite a bit. We'll fill it in tomorrow.

lib/models/task-dependency.js
```diff
@@ -0,0 +1,68 @@
...
+                defaultsTo: null
+            },
+            toJSON: function() {
+                // Remove waterline keys that we don't want in our dependency object
+                var obj = this.toObject();
+                delete obj.createdAt;
+                delete obj.updatedAt;
+                delete obj.id;
+                return obj;
+            }
```
tldavies: We dont care for a creation or update date?

benbp: We also do this in models.graphobjects. These are Waterline internal fields that we don't expose to code that accesses this data structure. `toJSON` is called before any documents are returned from waterline query methods.

lib/protocol/task-graph-runner.js
```diff
@@ -228,42 +228,30 @@ function taskGraphRunnerProtocolFactory (
         );
     };
 
-    TaskGraphRunnerProtocol.prototype.runTaskGraph = function (name, options, target) {
-        // TODO: BBP what asserts to we need here?
-        return messenger.request(
+    TaskGraphRunnerProtocol.prototype.runTaskGraph = function (graphId, domain) {
+        assert.uuid(graphId);
+        domain = domain || Constants.DefaultTaskDomain;
```
tldavies: Why is this a uuid assert while in other places it is a string assert for graphId?

lib/protocol/task-graph-runner.js
```diff
@@ -228,42 +228,30 @@ function taskGraphRunnerProtocolFactory (
...
-                }).then(function (result) {
-                    return message.resolve(
-                        new Result({ value: result })
-                    );
-                }).catch(function (error) {
-                    return message.reject(error);
-                });
-            }
+            'methods.runTaskGraph' + '.' + domain,
+            callback
```
tldavies: ;?

lib/protocol/task.js
```diff
@@ -19,49 +19,47 @@ function taskProtocolFactory (Promise, assert, Constants, Errors, messenger, _,
...
             Constants.Protocol.Exchanges.Task.Name,
-            'methods.cancel' + '.' + taskId,
+            'methods.cancel',
             function(data) {
+                if (!data.errName) {
+                    callback(data);
+                }
                 // Re-create error sent over the wire
                 if (Errors[data.errName]) {
                     callback(new Errors[data.errName](data.errMessage));
```
tldavies: do we need an else if data.errName is a not a valid error? Or will that never happen?

lib/workflow/messengers/messenger-AMQP.js
```diff
@@ -0,0 +1,76 @@
...
+
+module.exports = amqpMessengerFactory;
+amqpMessengerFactory.$provide = 'Task.Messengers.AMQP';
+amqpMessengerFactory.$inject = [
+    'Protocol.Task',
+    'Protocol.Events',
+    'Protocol.TaskGraphRunner',
+    'Promise'
+];
+function amqpMessengerFactory(
```
tldavies: newline

lib/workflow/messengers/messenger-AMQP.js
```diff
@@ -0,0 +1,76 @@
...
+
+    exports.subscribeRunTask = function(domain, callback) {
+        return taskProtocol.subscribeRun(domain, callback);
+    };
+
+    exports.publishRunTask = function(domain, taskId, graphId) {
+        return taskProtocol.run(domain, { taskId: taskId, graphId: graphId });
+    };
+
+    exports.subscribeCancel = function(callback) {
```
tldavies: should this be subscribeCancelTask?

lib/workflow/messengers/messenger-AMQP.js
```diff
@@ -0,0 +1,76 @@
...
+                domain,
+                taskId,
+                graphId,
+                state,
+                context,
+                terminalOnStates
+        );
+    };
+
+    exports.subscribeRunTaskGraph = function(domain, callback) {
```
tldavies: Does there need to be a publishRunTaskGraph to go along with this?

lib/workflow/stores/mongo.js
```diff
@@ -0,0 +1,563 @@
...
+mongoStoreFactory.$provide = 'TaskGraph.Stores.Mongo';
+mongoStoreFactory.$inject = [
+    'Services.Waterline',
+    'Promise',
+    'Constants',
+    'Errors',
+    'Assert',
+    '_'
+];
+function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
```
tldavies: newline

lib/protocol/task-graph-runner.js
benbp: Inconsistency, will update others to uuid.

lib/workflow/messengers/messenger-AMQP.js
benbp: ?

lib/workflow/messengers/messenger-AMQP.js
benbp: It could, the task protocol naming was this way for long lost reasons, so we were shooting for consistency in the short term.

lib/workflow/messengers/messenger-AMQP.js
benbp: Yeah, we don't leverage this event internally but someone could publish it externally via their own AMQP driver.

benbp: However, the new code updates might actually deprecate this, I'll investigate.

lib/workflow/stores/mongo.js
benbp: ?

lib/workflow/messengers/messenger-AMQP.js
tldavies: need a newline between the variables and the function

lib/workflow/stores/mongo.js
tldavies: need a newline between the variables and the function

benbp: ok

benbp:
And of course every PR I merge at this point will break this and need a rebase :)

lib/models/task-dependency.js
```diff
@@ -0,0 +1,68 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Copyright 2016

lib/models/task-dependency.js
yyscamper: why `id` is removed as well? isn't it a very useful query parameter to lookup database?

lib/models/task-events.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
```
yyscamper: miss copyright

lib/common/constants.js
```diff
@@ -157,14 +157,28 @@ function constantsFactory (path) {
...
             Succeeded: 'succeeded',
             Failed: 'failed',
             Cancelled: 'cancelled',
             Timeout: 'timeout',
-            Pending: 'pending'
+            Pending: 'pending',
+            Finished: 'finished'
         },
+        DefaultTaskDomain: 'default',
+        DefaultTaskLeaseAdjust: 2000,
```
yyscamper: It would be better if these Task constants are wrapped in a hierarchical structure just like other constants:
```javascript
Task: {
    DefaultDomain: 'default',
    DefaultLeaseAdjust: 2000,
    States: [...],
    FailedStates: [...],
    FinishedStates: [...]
}
```

lib/common/errors.js
```diff
@@ -82,6 +82,12 @@ function ErrorFactory(util) {
     }
     util.inherits(TaskCancellationError, BaseError);
 
+    function TaskStopError(message) {
```
yyscamper: What's the difference between TaskStopError and TaskCancellationError?

lib/protocol/task.js
```diff
@@ -19,49 +19,47 @@ function taskProtocolFactory (Promise, assert, Constants, Errors, messenger, _,
...
+    TaskProtocol.prototype.subscribeCancel = function(callback) {
         assert.func(callback);
 
         return messenger.subscribe(
             Constants.Protocol.Exchanges.Task.Name,
-            'methods.cancel' + '.' + taskId,
+            'methods.cancel',
             function(data) {
+                if (!data.errName) {
+                    callback(data);
```
yyscamper: Seems there should be a `return` here

lib/workflow/messengers/messenger-AMQP.js
```diff
@@ -0,0 +1,76 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Copyright 2016

lib/workflow/messengers/messenger-AMQP.js
```diff
@@ -0,0 +1,76 @@
...
+    exports.publishCancelTask = function(taskId, errName, errMessage) {
+        return taskProtocol.cancel(taskId, errName, errMessage);
+    };
+
+    exports.subscribeTaskFinished = function(domain, callback) {
+        return eventsProtocol.subscribeTaskFinished(domain, callback);
+    };
+
+    exports.publishTaskFinished = function(
+            domain,
```
yyscamper: incorrect indent.

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+    function TaskGraph(definition, context, domain) {
+        this.definition = definition;
+
+        if (this.definition.options && this.definition.options.instanceId) {
+            this.instanceId = this.definition.options.instanceId;
+        } else {
+            this.instanceId = uuid.v4();
+        }
+
+        // Bool
```
yyscamper: What does this comment means?

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+        this.context.graphId = this.instanceId;
+        this.domain = domain || Constants.DefaultTaskDomain;
+        this.name = this.definition.friendlyName;
+        this.injectableName = this.definition.injectableName;
+
+        this.tasks = {};
+
+        // TODO: find instances of 'valid' elsewhere and replace from valid to Pending
+        // TODO: replace _status with status
+        this._status = Constants.TaskStates.Pending;
```
yyscamper: Why only this variable `_status` begins with an underscore? We have to to this?

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+            });
+            var taskOverrides = {
+                instanceId: idMap[taskData.label],
+                waitingOn: taskData.waitOn,
+                ignoreFailure: taskData.ignoreFailure
+            };
+
+            if (taskData.taskName) {
+                return self.constructTaskObject(taskData.taskName, taskOverrides,
+                        taskData.optionOverrides, taskData.label)
```
yyscamper: incorrect indent.

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+        definition.name = taskOverrides.name || definition.injectableName;
+        definition.waitingOn = taskOverrides.waitingOn || {};
+        // TODO: Remove ignoreFailure in favor of better graph branching evalution.
+        // NOTE: actually ignoreFailure is still useful for tasks that can
+        // fail but we don't care AND don't have anything queued up to run
+        // on failure (rare case probably, but still worth supporting IMO).
+        definition.ignoreFailure = taskOverrides.ignoreFailure || false;
+
+        var allOptions = _.uniq(
+                _.keys(definition.options).concat(baseTaskDefinition.requiredOptions));
```
yyscamper: incorrect indent.

spec/lib/workflow/messengers/messenger-amqp-spec.js
```diff
@@ -0,0 +1,132 @@
...
+        var callback = function() {};
+        return amqp.subscribeRunTaskGraph('default', callback)
+        .then(function() {
+            expect(taskGraphRunnerProtocol.subscribeRunTaskGraph).to.have.been.calledOnce;
+            expect(taskGraphRunnerProtocol.subscribeRunTaskGraph)
+                .to.have.been.calledWith('default', callback);
+        });
+    });
+
+    it('subscribeCancelGraph');
```
yyscamper: Forgot this?

yyscamper:
Just quickly go through the code and find some style error, need more time for the detail design review. Since I'm in vacation, so the detail review cannot finish in a short time.If other reviewer agree, you can go ahead the merge.

jlongever:
test this please

lib/models/task-dependency.js
benbp: Historically we have removed the id field because it gets annoying in logging and data traversal since it's actually a BSON object, and it also can get confusing to newcomers to the code when trying to figure out the difference between an document id, an instanceId, a nodeId, etc. Also, often all we need is the instanceId for queries/internal logic and so it's unnecessary to have the id field. In this refactor, in the cases where we do want to lookup by a document ID, we are already for other reasons using the waterline mongo native method so the toJSON call gets bypassed.

lib/models/task-dependency.js
benbp: Thanks, I new I was missing some.

lib/common/constants.js
benbp: Good idea. We'll make this update.

lib/common/errors.js
benbp: A TaskCancellationError is an error caused by an external cancellation event that triggers downstream graph handling logic. A TaskStopError is a special error type only used internally to resolve conflicts when multiple task runners pick up the same task because of a heartbeat failure of one of the runners. In this case, we want to stop running the duplicate task but not persist any state or publish any events, since that will cause the graph to re-evaluate itself and potentially fail.

benbp: @VulpesArtificem let's add a comment either here or in the Task.js code elaborating on this.

lib/protocol/task.js
benbp: Yep, that would be much better to exit the function here. It looks like we still won't evaluate the `if` statement below, so at least there's not an existing bug.

lib/workflow/task-graph.js
benbp: Just a simple type hint. `this.definition.serviceGraph` is a boolean value.

lib/workflow/task-graph.js
benbp: It's a terrible variable name that I chose like a year ago. I'd like to change it, but I also don't want to break backwards compatibility. My thought is we could add another value "state" and start using both, then eventually deprecate "_status".

lib/workflow/task-graph.js
benbp: How would you prefer we indent in this case, like this maybe?

```
return self.constructTaskObject(
                taskData.taskName, taskOverrides, taskData.optionOverrides, taskData.label)
```

benbp: Oh, or you mean:

```
return self.constructTaskObject(taskData.taskName, taskOverrides,
    taskData.optionOverrides, taskData.label)
```

benbp: As a personal style preference, I like to keep function arguments further indented away from the part of the call that doesn't include the function name, just because it's clearer to me when reading that they are arguments to the previous line, and not a new line of logic or anything. I acknowledge that it's personal preference and am curious what others prefer in these cases. I don't think we have any sort of standardized style rules within our codebase pertaining to this corner case, but do you think we would benefit from it?

lib/workflow/task-graph.js
benbp: Similar reasoning to my above comment, though I'll add an emphasis on how the extra indentation here makes it easier to see/read the variable name by itself as one is scanning through the code. Happy to change it though, just adding some background on why my preferences are they way they are.

spec/lib/workflow/messengers/messenger-amqp-spec.js
benbp: @VulpesArtificem ?

benbp:
Thanks for the style/minor code errors @yyscamper. We'll go ahead with the merge if we get enough reviews before you are back, and I'll make sure to reach out to you/address any larger design comments you have when you return. Have a good vacation!

lib/workflow/stores/mongo.js
```diff
@@ -0,0 +1,563 @@
...
+
+    exports.findActiveGraphs = function(domain) {
+        assert.string(domain, 'domain');
+
+        var query = {
+            domain: domain,
+            _status: Constants.TaskStates.Pending
+        };
+
+        return waterline.graphobjects.find(query);
```
tldavies: should this be findMongo?

benbp: Nope. We only pass through to the native calls in places where we need to use the mongo query syntax/response in a way that's not supported by the waterline ORM layer.

benbp: There are some conveniences to the response objects that come out of waterline too, so we like to use them where we can.

lib/workflow/stores/mongo.js
```diff
@@ -0,0 +1,563 @@
...
+                $lt: new Date(Date.now() - leaseAdjust)
+            },
+            state: Constants.TaskStates.Pending
+        };
+
+        return waterline.taskdependencies.find(query);
+    };
+
+    exports.expireLease = function(objId) {
+        assert.string(objId, 'objId');
```
tldavies: is objId a string or a uuid?

lib/workflow/stores/store.js
```diff
@@ -0,0 +1,18 @@
+// Copyright 2015, EMC, Inc.
+'use strict';
+
+module.exports = storeFactory;
+storeFactory.$provide = 'TaskGraph.Store';
+storeFactory.$inject = [
+    'Services.Configuration',
+    '$injector'
+];
+function storeFactory(configuration, injector) {
```
tldavies: newline

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+                taskOverrides, label);
+    };
+
+    TaskGraph.prototype.constructTaskObject = function(taskName, taskOverrides,
+           optionOverrides, label) {
+        var self = this;
+        return store.getTaskDefinition(taskName)
+        .then(function(taskDefinition) {
+            return self._buildTaskDefinition(taskDefinition,
+                    optionOverrides, taskOverrides, label);
```
tldavies: is the indent off?

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+        // Update shared context with properties from this task
+        var _properties = _.merge(taskDefinition.properties, baseTaskDefinition.properties);
+        context.properties = _.merge(_properties, context.properties);
+    };
+
+    TaskGraph.prototype._validateOptions = function(taskDefinition, label) {
+        var self = this;
+        var baseTaskDefinition = self._getBaseTask(taskDefinition);
+        _.forEach(baseTaskDefinition.requiredOptions, function(k) {
+            var option = taskDefinition.options[k];
```
tldavies: Whats the k stand for?

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+        var self = this;
+
+        // nested key is a dot notated string that represents a JSON scope, e.g.
+        // os.linux.type represents { os: { linux: { type: 'value' } } }
+        if (!nestedKey) {
+            return;
+        }
+        assert.string(nestedKey);
+        var keys = nestedKey.split('.');
+        if (keys.length === 1) {
```
tldavies: newline

lib/workflow/stores/mongo.js
benbp: It's a stringified BSON id, so a string but not a UUID.

lib/workflow/task-graph.js
benbp: key.

benbp: In this case we're inspecting the current options object for the keys specified in the requiredOptions object.

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+        .then(function() {
+            return self;
+        });
+    };
+
+    TaskGraph.create = function (domain, data) {
+        var definition = data.definition;
+        var options = data.options;
+        var context = data.context;
+        // TODO: replace this with _.defaultsDeep when we upgrade to lodash 3 so
```
tldavies: we upgraded to lodash 3. Should this updated?

benbp: Aw hell yeah!

spec/lib/workflow/messengers/messenger-amqp-spec.js
```diff
@@ -0,0 +1,132 @@
...
+        });
+        _.forEach(eventsProtocol, function(method) {
+            method.reset();
+        });
+        _.forEach(taskGraphRunnerProtocol, function(method) {
+            method.reset();
+        });
+    });
+
+    it('subscribeRunTask', function() {
```
tldavies: it need to start with a "should"

spec/lib/workflow/messengers/messenger-amqp-spec.js
tldavies: need to add tests.

spec/lib/workflow/messengers/messenger-amqp-spec.js
```diff
@@ -0,0 +1,132 @@
...
+            expect(taskGraphRunnerProtocol.subscribeRunTaskGraph)
+                .to.have.been.calledWith('default', callback);
+        });
+    });
+
+    it('subscribeCancelGraph');
+
+    it('publishCancelGraph');
+
+    it('start', function() {
```
tldavies: i really dont like this one. start is way to generic.

spec/lib/workflow/task-graph-spec.js
```diff
@@ -0,0 +1,469 @@
...
+                definition: definitions.graphDefinition,
+                options: { nodeId: 'testnode' }
+            })
+            .then(function(graph) {
+                expect(graph).to.have.property('node').that.equals('testnode');
+            });
+        });
+
+        it('should create a graph with an inline task definition', function() {
+            return TaskGraph.create('domain', { definition: definitions.graphDefinitionInline });
```
tldavies: do we need an expect here?

tldavies:
Looks good. Need to increase your code coverage though.

lib/common/constants.js
heckj: :+1:

lib/workflow/stores/mongo.js
```diff
@@ -0,0 +1,563 @@
...
+    'Assert',
+    '_'
+];
+function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
+    var exports = {};
+
+    // NOTE: This is meant to be idempotent, and just drop the update silently
+    // if the graph has already been marked as done elsewhere and the query returns
+    // empty.
+    exports.setGraphDone = function(state, data) {
```
heckj: since this appears to be the primary (and only) interface point into the backend stores for the workflow atomicity, could you add JSDoc to these `exports` methods that describe the inputs and return values expected

lib/workflow/task-graph.js
```diff
@@ -0,0 +1,493 @@
...
+function taskGraphFactory(
+    taskLibrary,
+    store,
+    Constants,
+    assert,
+    uuid,
+    Promise,
+    _
+) {
+    function TaskGraph(definition, context, domain) {
```
heckj: would you JSDoc the constructor details here? Would make this a lot more understandable down the road - what's in definition, context, and domain...

heckj:
:+1: with add'l JSDoc entries and the other comments above.

Seems like some of this code would benefit from tests that didn't mock the waterline later and leverage the store right down to mongo, running against an existing mongo instance to make sure we're getting the responses we expect so that updates in mongo drivers or later versions don't mess us up unexpectedly.

benbp:
Will do @heckj. @jlongever FYI I think this might be a good time to bring back up our discussion about where to put/what we want to do with node.js-based tests that access internal interfaces, but with external dependencies like mongo running (basically our system vs. isolated tests thing) (see Joe's comment above).

jlongever:
@benbp @heckj  Personally I'd prefer to keep external service dependencies out of unit-test code and develop functional integration tests. Fwiw,  there has been a unit-test refactoring effort happening to remove those external dependencies.  @amymullins fyi

benbp:
@jlongever yeah agreed we want to keep these service deps out of the main unit test suite, though I was alluding to the potential need for functional integration tests that act like unit tests essentially (i.e. written closer to actual code, not behind an API black box, but with external service dependencies in the test environment).

benbp:
This was our whole discussion when I was in MA before anyone had done any of the integration test implementation about having both the black box python-driven suite and a node driven suite (system as you called it?), because they both filled different needs.

jlongever:
@benbp Yeah, definitely.  I'd want to define it around it's own functional test framework (as opposed to munging in with SW unit-tests) but I'm open to building up a framework to cover those cases.

lib/models/graph-object.js
```diff
@@ -38,8 +38,13 @@ function GraphModelFactory (Model) {
             node: {
                 model: 'nodes'
             },
-            deserialize: function() {
-                // Remove waterline keys that we don't want in our graph objects
+            /*
+            runtime: {
```
jlongever: remove unused code?

lib/workflow/stores/mongo.js
jlongever: Hey @benbp Is there any plan to deprecate waterline completely and apply consistent abstraction methods (just native mongo)?  I can see this helping with general maintainability and reduction in complexity/usage (at least from a outside developer looking in)... plus any behavioral differences.  Do those conveniences outweigh the benefits of using mongo driver exclusively?

lib/workflow/task-graph.js
VulpesArtificem: ?

benbp: I think she means a newline above the if statement.

zyoung51: Could lodash .get be used here?  Assert (value === _.get (obj, nestedKey, null))

lib/workflow/task-graph.js
yyscamper: From your explanation, this sure brings some benefit to distinguish longer argument or longer logic. I just searched some popular JS style guide, they don't have such a style definition, they always prefer the 4-space indention in nearly every case (the exception case is for argument vertical alignment and `? :`)

What's more, I'm curious how the VIM automatically does this kind of "smart" indentation,  we have to do it manually?

jlongever:
resolve conflicts, more unit-tests.. otherwise +1

jlongever:
test this please

PR #78, opened 2016-02-05
Force util.isError to return true for SchemaErrors

brianparry:
This change is required to get the Swagger error handler to parse SchemaErrors properly (see https://github.com/RackHD/on-http/pull/119)

Swagger runs its error parser when next is called with a non-null error argument.  The error parser then uses util.isError to determine if the object is an Error.  Since util.isError only returns true if Object.prototype.toString.call(err) returns '[object Error]', util.isError returns false when it is passed an instance of jsonschema.SchemaError.  My solution is to wrap the SchemaError object in an Error object.

I'm not convinced this is the best way to solve this problem, so I'm open to suggestions.

### DIFF
--- a/lib/common/validatable.js
+++ b/lib/common/validatable.js
@@ -43,7 +43,7 @@ function ValidatableFactory (Promise, _, jsonschema, Errors, assert) {
         if (_.isEmpty(outcome.errors)) {
             return Promise.resolve(target);
         } else {
-            return Promise.reject(new Errors.SchemaError(outcome));
+            return Promise.reject(new Error(new Errors.SchemaError(outcome)));
         }
     };
 
--- a/spec/lib/common/validatable-spec.js
+++ b/spec/lib/common/validatable-spec.js
@@ -47,7 +47,7 @@ describe('Validatable', function () {
             it('should reject on failure', function () {
                 return new Subject(
                     { name: 1337 }
-                ).validate().should.be.rejectedWith(Errors.SchemaError);
+                ).validate().should.be.rejectedWith(Error);
             });
         });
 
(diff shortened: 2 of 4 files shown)


benbp:
I'm wondering with @rolandpoulter whether there's a way that we update our own error classes to respond to this call with '[object Error]' instead?

benbp:
@rolandpoulter and I were able to get BaseError responding appropriately but still trying to figure out how to extend it up to the subclasses:

```
 17     function BaseError(message, context) {
 18         // this.message = message;
 19         // this.name = this.constructor.name;
 20         // this.context = context || {};
 21         var self = new Error(message);
 22         self.name = this.constructor.name;
 23         self.context = context || {};
 24         self.__proto__ = this.constructor.prototype; /* jshint ignore:line */
 25         Error.captureStackTrace(self, this.constructor);
 26
 27         return self;
 28     }
 29     util.inherits(BaseError, Error);
```

A little more background that we found is here: https://github.com/nodejs/node-v0.x-archive/issues/3212

benbp:
Just to clarify, I don't want to block this, so :+1:, but it would be cool if we could figure out a way to actually make our error objects appear to be real error objects at the deepest level, even though javascript makes it really hard for us, I'm sure we're not the first.

jlongever:
+1

brianparry:
Thanks @benbp.  I totally agree that the ideal way to do this is to make our error objects appear to be real error objects.  I actually fiddled with this for a couple hours, but wasn't able to come up with anything better than this.  util.isError is a very strict test.  Not only does it use toString but it uses toString from Object.prototype so its behavior can't be overridden.

benbp:
Yeah, maybe it's not worth the effort in this case. Though it would be nice to add a little more specificity to the tests, perhaps some checking against the error message, so we know we're causing the error we think we're causing?

:+1: still though, do with it what you will :)

PR #79, opened 2016-02-06
Switch Dockerfile to Alpine Linux

rolandpoulter:
* Switched to an alpine linux base image with node v4 installed. 
* Install git on base image.
* Added a error message when configuration file is missing.

### DIFF
--- a/lib/services/configuration.js
+++ b/lib/services/configuration.js
@@ -44,8 +44,8 @@ function configurationServiceFactory(
             //TODO: remove once OnRack build process adjusts for changes.
             nconf.file('global', Constants.Configuration.Files.OnRack);
         } else {
-          console.error('Failed to load configuration file:',
-            Constants.Configuration.Files.Global);
+            console.error('Failed to load configuration file:',
+                Constants.Configuration.Files.Global);
         }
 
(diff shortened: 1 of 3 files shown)


lib/services/configuration.js
```diff
@@ -43,6 +43,9 @@ function configurationServiceFactory(
         } else if (fs.existsSync(Constants.Configuration.Files.OnRack)) {
             //TODO: remove once OnRack build process adjusts for changes.
             nconf.file('global', Constants.Configuration.Files.OnRack);
+        } else {
+          console.error('Failed to load configuration file:',
+            Constants.Configuration.Files.Global);
```
benbp: The logger is injected here, can we use that?

rolandpoulter: I don't think so. Possibly, I think the logger need certain configuration properties before it will work. I tried using it here when I was debugging and it was failing.

I added that when I was trying to mount `/opt/monorail/config.json` for RackHD docker containers. Took me a while to realize the config was missing. This made my life much easier.

rolandpoulter: Although it might be better as a `console.warn`?

VulpesArtificem:
:+1:

benbp:
:+1:

PR #81, opened 2016-02-08
On-core unit test re-factor to remove RabbitMQ dependency

uppalk1:
# refactor connection-spec.js

### DIFF
--- a/spec/lib/common/connection-spec.js
+++ b/spec/lib/common/connection-spec.js
@@ -1,3 +1,4 @@
+
 // Copyright 2015, EMC, Inc.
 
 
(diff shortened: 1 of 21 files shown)


spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
+                        'on': function (a, callback) {
+                            if (a === 'error') {
+                                callback({'message': 'error'}, {});
+                            }
+                        }
+                    })
+                })
+                return this.subject.start()
+                    .catch(function (Error) {
+                        console.log(Error);
```
benbp: I don't think we would want to catch here, otherwise the test will pass on start() failures.

spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
-        describe('exchanges', function () {
-            it('should be undefined if not connected', function () {
-                expect(this.subject.exchanges).equal(undefined);
-            });
+            describe('exchange', function ()  {
+                it('should not provide an exchange if connection not established ', function () {
+                    var self = this;
+                    return this.subject.exchange()
+                        .catch(function (Error) {
+                            console.log(Error);
```
benbp: Same as above, I don't think we want to catch here because it will prevent the test from failing.

spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
+                                callback(null, {});
+                            },
+                            'exchange': function (a, b, callback){
+                                callback('Hello', {})
+                            },
+                            'exchanges':
+                                'on.test'
+                        })
+                    })
+                    this.subject.start().then(function () {
```
benbp: This promise should be returned so that the test runner can catch errors.

spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
+                        if (self.subject.connected == true) {
```
benbp: Do we want a failure case if `self.subject.connected` does not equal true? Otherwise it's a noop test in this case.

spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
+                            var name = 'test';
+                            return self.subject.exchange(name)
+                                .catch(function (Error) {
```
benbp: Same as above, probably don't want to catch here.

spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
+                            },
+                            'exchange': function (a, b, callback){
+                                callback('exchange', {})
+                            },
+                            'exchanges':
+                                'on.test'
+                        })
+                    })
+                    this.subject.start().then(function () {
+                        if (self.subject.connected == true) {
```
benbp: Same as comments above re: handling when connected !== true.

spec/lib/common/connection-spec.js
```diff
@@ -5,113 +6,254 @@
...
+                            var name = 2;
+                            return self.subject.exchange(name)
+                            callback('exchange', {});
```
benbp: Is this ever going to get called if we return above?

jlongever:
Looks like you pulled in a bunch of other commits,  could you rebase with just your changes?

uppalk1:
Closing this PR due to merge issues... will reopen...

PR #82, opened 2016-02-09
Add environment collection with hierarchy support

zyoung51:
- Allow the user to establish arbitrarily named environment
  collections and merge/retrieve items from those arbitrary
  environments in a priority order.

### DIFF
--- a/lib/services/environment.js
+++ b/lib/services/environment.js
@@ -1,0 +1,83 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = environmentServiceFactory;
+
+environmentServiceFactory.$provide = 'Services.Environment';
+environmentServiceFactory.$inject = [
+    'Promise',
+    'Services.Waterline',
+    '_',
+    'Assert'
+];
+
+function environmentServiceFactory(
+    Promise,
+    waterline,
+    _,
+    assert
+) {
+    function EnvironmentService() {
+    }
+
+    /**
+     * Set the 'key' to the 'value' in the document identified by 'identifier'
+     * @param  {String}     key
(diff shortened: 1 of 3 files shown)


zyoung51:
@RackHD/corecommitters @uppalk1 @keedya @tannoa2

lib/services/environment.js
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    /**
+     * Retrieve the documents specified by identifiers and merge in order of priority
+     * @param  {Array}   identifiers
+     */
+    EnvironmentService.prototype.getAll = function (identifiers) {
+        identifiers = identifiers || ['global'];
+        assert.arrayOfString(identifiers, 'identifiers should be an array');
+        return Promise.all(Promise.map( identifiers, function(identifier) {
```
benbp: Is the `all` necessary?

benbp: Or does `.map` hit a spread normally? Guess I'm forgetting this behavior.

benbp:
:+1: on the code, but perhaps name/document this to be a little more descriptive. If this applies only to SKUs at the moment as I think it does, then the naming is kind of abstract in that regard and without any context I think it might be hard to tell what this is actually used for.

If the intent is to make this just a generic collection that could be utilized in many ways, I'm all for it, but I still think a nod to "this is used by SKUs, but could also be used for things like X" in the comments would be good.

zyoung51:
@benbp, the goal here was to make something generic that could be used for more, but I personally only plan to use it in skupacks to feed option/configuration values from the skupack config.json into templates during rendering.

jlongever:
+1

PR #83, opened 2016-02-10
on-core refactor unit tests

uppalk1:


spec/lib/common/connection-spec.js
```diff
@@ -5,112 +6,255 @@
...
+                            },
+                            'exchange': function (a, b, callback){
+                                callback('Hello');
+                            },
+                            'exchanges':
+                                'on.test'
+                        });
+                    });
+                    return this.subject.start().then(function () {
+                        if (self.subject.connected === true) {
```
benbp: Do we want to fail this test if connected !== true? Otherwise I'm worried about the potential for this test producing a false positive as code changes in the future.

spec/lib/common/connection-spec.js
```diff
@@ -5,112 +6,255 @@
...
+                            },
+                            'exchange': function (a, b, callback){
+                                callback('exchange', {});
+                            },
+                            'exchanges':
+                                'on.test'
+                        });
+                    });
+                    return this.subject.start().then(function () {
+                        if (self.subject.connected === true) {
```
benbp: Same as above comment.

spec/lib/common/connection-spec.js
```diff
@@ -5,112 +6,255 @@
...
+                            'on': function (a, callback) {
+                                if(a === 'ready'){
+                                    callback(null, {});
+                                }
+                            },
+                            'disconnect':function(){done();},
+                        });
+                    });
+                    return this.subject.start().then(function () {
+                        return expect(self.subject.stop()).to.be.ok;
```
benbp: Does `stop()` return a Promise? If so, we'll want to assert `return expect(self.subject.stop()).to.be.fulfilled` because I think `.to.be.ok` will pass even for an error object since it just detects a truthy/falsey condition.

benbp:
A couple comments, and the lint errors should be addressed. Also, before the merge you'll want to git rebase this branch against master to get rid of that merge commit you have ("Merge pull request #1 from RackHD/master").

uppalk1:
test this please

uppalk1:
test this please

jlongever:
test this please

uppalk1:
test this please

jlongever:
test this please

jlongever:
+1

tannoa2:
+1

PR #84, opened 2016-02-10
Add Profile scoping support for skupacks

zyoung51:
- Refactor Template/Profiles to combine scoping support logic for
  both into DbRenderableContent.

### DIFF
--- a/lib/common/db-renderable-content.js
+++ b/lib/common/db-renderable-content.js
@@ -38,7 +38,7 @@ function dbRenderableContentServiceFactory(
     var logger = Logger.initialize(dbRenderableContentServiceFactory);
 
     function DbRenderableContentService (options) {
-        options = options || {}
+        options = options || {};
         this.loader = new FileLoader();
         this.collectionName = options.collectionName;
         this.directory = options.directory;
--- a/lib/common/db-renderable-content.js
+++ b/lib/common/db-renderable-content.js
@@ -1,0 +1,123 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = dbRenderableContentServiceFactory;
+
+dbRenderableContentServiceFactory.$provide = 'DbRenderableContent';
+dbRenderableContentServiceFactory.$inject = [
+    'FileLoader',
+    'Constants',
+    'Promise',
+    '_',
+    'Services.Waterline',
+    'Logger',
+    'ejs'
+];
+
+/**
+ * dbRenderableContentServiceFactory provides the template-service singleton object.
+ * @private
+ * @param  {FileLoader} FileLoader A FileLoader class for extension.
+ * @param  {configuration} configuration      An instance of the configuration configuration object.
+ * @return {TemplateService}            An instance of the TemplateService.
+ */
+function dbRenderableContentServiceFactory(
+    FileLoader, 
+    Constants, 
+    Promise, 
+    _, 
+    waterline, 
+    Logger, 
+    ejs ) 
+{
+    var logger = Logger.initialize(dbRenderableContentServiceFactory);
+
+    function DbRenderableContentService () {
+        this.loader = new FileLoader();
+    }
+
+    DbRenderableContentService.prototype.load = function load() {
+        var self = this;
+
+        return this.loader.getAll(
+            self.directory
+        ).then(function (templates) {
+            var promises = _.map(templates, function (contents, name) {
+                return self.put(name, contents.toString());
+            });
+
+            return Promise.all(promises);
+        }).catch(function (err) {
+            if (process.env.NODE_ENV !== 'test') {
+                logger.error(
+                    'Unable to load ' + self.collectionName + ' from ' + self.directory + '.',
+                    {
+                        error: err
+                    }
+                );
+            }
+        });
+    };
+
+    DbRenderableContentService.prototype.put = function (file, contents, scope) {
+        var self = this;
+        scope = scope || Constants.Scope.Global;
+        return Promise.resolve(
+            waterline[self.collectionName].findOne({ name: file, scope: scope })
+            .then(function(doc) {
+                if (!_.isEmpty(doc)) {
+                    return waterline[self.collectionName].update(
+                        { name: file, scope: scope },
+                        { contents :contents }
+                    );
+                } else {
+                    return waterline[self.collectionName].create(
+                        {
+                            name: file,
+                            contents: contents,
+                            scope: scope
+                        }
+                    );
+                }
+            })
+        );
+    };
+
+    DbRenderableContentService.prototype.get = function (name, scope) {
+        var self = this;
+        scope = scope || [ Constants.Scope.Global ];
+
+        // The position of the tag defines the scope priority
+        var scopeWeight = {}, i = 1;
+        _.forEach(scope, function(item) {
+            scopeWeight[item] = i;
+            i += 1;
+        });
+
+        return waterline[self.collectionName].find({name: name, scope: scope})
+            .then(function(templates) {
+                templates.sort(function(a,b) {
+                    return scopeWeight[a.scope] - scopeWeight[b.scope];
+                });
+                return templates[0];
+            });
+    };
+
+    DbRenderableContentService.prototype.getAll = function () {
+        return waterline[this.collectionName].find();
+    };
+
+    DbRenderableContentService.prototype.render = function render (name, options, scope) {
+        return this.get(name, scope).then(function (template) {
+            return ejs.render(template.contents, options);
+        });
+    };
+
+    DbRenderableContentService.prototype.unlink = function (name, scope) {
+        scope = scope || Constants.Scope.Global;
+        return waterline[this.collectionName].destroy({name: name, scope: scope});
+    };
+
+    return DbRenderableContentService;
+}
--- a/lib/common/db-renderable-content.js
+++ b/lib/common/db-renderable-content.js
@@ -18,9 +18,13 @@ dbRenderableContentServiceFactory.$inject = [
 /**
  * dbRenderableContentServiceFactory provides the template-service singleton object.
  * @private
- * @param  {FileLoader} FileLoader A FileLoader class for extension.
- * @param  {configuration} configuration      An instance of the configuration configuration object.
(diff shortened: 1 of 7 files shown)


lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+        return this.loader.getAll(
+            self.directory
+        ).then(function (templates) {
+            var promises = _.map(templates, function (contents, name) {
+                return self.put(name, contents.toString());
+            });
+
+            return Promise.all(promises);
+        }).catch(function (err) {
+            if (process.env.NODE_ENV !== 'test') {
```
benbp: Are we using this convention elsewhere? If so, might be good for us to add a quick note in the README or something.

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+                        error: err
+                    }
+                );
+            }
+        });
+    };
+
+    DbRenderableContentService.prototype.put = function (file, contents, scope) {
+        var self = this;
+        scope = scope || 'global';
```
benbp: Do you think it's a good idea to throw 'global' into Constants at this point?

zyoung51: Maybe after typing it a thousand more times.... 
Yeah, I should add that to Constants.

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+        return Promise.resolve(
+            waterline[self.collectionName].findOne({ name: file, scope: scope })
```
benbp: Are these waterline methods still not returning native bluebird promises on your end? If so, we should file a bug.

zyoung51: This is a copy/paste out of templates.js: https://github.com/RackHD/on-core/pull/84/files#diff-1c2fcbad07fcbf53741831836df8054dL60 and profile.js: https://github.com/RackHD/on-core/pull/84/files#diff-c4b2d9b82fc1aaf845aef6364f13f1b4L61

lib/common/db-renderable-content.js
benbp: Lol

lib/common/db-renderable-content.js
benbp: Ah ok.

lib/common/profile.js
```diff
@@ -30,66 +36,20 @@ function profileServiceFactory(FileLoader, Constants, Promise, _, waterline, Log
      * @extends {FileLoader}
      */
     function ProfileService () {
-        this.loader = new FileLoader();
+        dbRenderable.call(this);
+        this.directory = Constants.Profiles.Directory;
+        this.collectionName = 'profiles';
```
benbp: Part of me is wondering whether we would be better off with something like `waterline.profiles.identify` rather than a magic collection string, but part of me is reluctant to include that as a dependency at the same time.

lib/common/profile.js
```diff
@@ -21,8 +20,15 @@ profileServiceFactory.$inject = [
  * @param  {configuration} configuration      An instance of the configuration configuration object.
  * @return {ProfileService}            An instance of the ProfileService.
  */
-function profileServiceFactory(FileLoader, Constants, Promise, _, waterline, Logger) {
-    var logger = Logger.initialize(profileServiceFactory);
+function profileServiceFactory(
+    Constants, 
+    Promise, 
+    _, 
+    dbRenderable,
```
benbp: Caps here (`DbRenderable`) following javascript class convention?

benbp:
A couple of minor nits, I think it would also be good to add some simple unit tests at the DbRenderableContent level, even if they are exercised 100% by the profiles/templates unit tests. :+1: conditional on that.

I like the DbRenderableContent abstraction.

zyoung51:
Thanks @benbp, I'll update the PR.  Hopefully adding the DbRenderableContent unittests gets me that 0.002% coverage back: Coverage decreased (-0.002%) to 94.86% :-)

benbp:
Lol, yeah how dare you, it's really unacceptable.

lib/common/profile.js
zyoung51: Moved the waterline dependency from here to db-renderable, and reluctant to re-include it for the name.  Maybe a constant should be used here and in the models?

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+    '_',
+    'Services.Waterline',
+    'Logger',
+    'ejs'
+];
+
+/**
+ * dbRenderableContentServiceFactory provides the template-service singleton object.
+ * @private
+ * @param  {FileLoader} FileLoader A FileLoader class for extension.
```
yyscamper: The `{}` should be filled with the argument type, not name, such as
```
@param {Object} FileLoader - A FileLoader class for extension.
```

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+                );
+            }
+        });
+    };
+
+    DbRenderableContentService.prototype.put = function (file, contents, scope) {
+        var self = this;
+        scope = scope || Constants.Scope.Global;
+        return Promise.resolve(
+            waterline[self.collectionName].findOne({ name: file, scope: scope })
```
yyscamper: The `Promise.resolve` is redundant and should be removed, as `waterline.findOne` will return a Promise.

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+            return ejs.render(template.contents, options);
+        });
+    };
+
+    DbRenderableContentService.prototype.unlink = function (name, scope) {
+        scope = scope || Constants.Scope.Global;
+        return waterline[this.collectionName].destroy({name: name, scope: scope});
+    };
+
+    return DbRenderableContentService;
```
yyscamper: Just want to confirm, what you need is a `Class` not a `Instance`?

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+        scope = scope || [ Constants.Scope.Global ];
+
+        // The position of the tag defines the scope priority
+        var scopeWeight = {}, i = 1;
+        _.forEach(scope, function(item) {
+            scopeWeight[item] = i;
+            i += 1;
+        });
+
+        return waterline[self.collectionName].find({name: name, scope: scope})
```
yyscamper: How `self.collectionName` comes from? Is this line (https://github.com/RackHD/on-core/pull/84/files#diff-c4b2d9b82fc1aaf845aef6364f13f1b4R39) ?

This kind of usage is a little tricky for me, it is not so easy and intuitive to use. I'd rather pass an argument in constructor function or the `load` function.

zyoung51: The intention is to never use this object directly.  This object is meant to be inherited and used by another object.

lib/common/db-renderable-content.js
zyoung51: Confirmed..

spec/lib/common/db-renderable-content-spec.js
```diff
@@ -0,0 +1,254 @@
...
+        it('should log warnings on load failure', function() {
+            process.env.NODE_ENV = 'not test';
+            return this.subject.load()
+            .then(function() {
+                expect(Logger.prototype.log).to.have.been.calledWithMatch(
+                    'error', /Unable to load things/, { error: new Error() });
+            });
+        });
+
+        it('should not log warnings in test env', function() {
```
yyscamper: Why the test env doesn't expect the warnings information?

lib/common/db-renderable-content.js
```diff
@@ -0,0 +1,123 @@
...
+                            contents: contents,
+                            scope: scope
+                        }
+                    );
+                }
+            })
+        );
+    };
+
+    DbRenderableContentService.prototype.get = function (name, scope) {
```
yyscamper: I'd expect add some JSDoc for this function, this function contains some priority ranking, need some special description for it.

spec/lib/common/db-renderable-content-spec.js
zyoung51: Copy/pasted from legacy code...  @benbp @heckj Do you remember why the code was written this way?

lib/common/db-renderable-content.js
yyscamper: If for inheritance, I think we'd better define this common property in base class.

spec/lib/common/db-renderable-content-spec.js
heckj: Heh, legacy is all about John's setting up for the globing/loading in the on-core libraries if I remember. I think he set up this pattern to not have to see the log messages in running tests (which where a combo of unit and functional) where they "weren't useful", as at the time we have logging extremely tightly tied to the messaging layer rather than as a layer where it could shim in as needed. 

My own opinion is that we could easily change this - as long as the warning message really is a warning and not set at the wrong level, I personally kinda like to see them, even in log output from unit tests

lib/common/db-renderable-content.js
heckj: @benbp I think it's a copy-pasta from legacy patterns that John set up about not wanting to see his own warnings in unit tests. Don't know if it's relevant any longer, may be exactly the opposite.

yyscamper:
@zyoung51 , the new houndci warning "Missing semicolon" should be fixed.

heckj:
Don't know what (if anything) you want to do about the "don't log warning msgs if ENV==test", but good to merge in either case.

:+1:

lib/common/db-renderable-content.js
benbp: I bet we can get rid of it.

spec/lib/common/db-renderable-content-spec.js
benbp: Yeah, I believe we've since added logging mocks everywhere anyways.

benbp:
@yyscamper are you a -1? 

:+1: from me, pending Felix.

amymullins:
@yyscamper , any additional feedback on this ?  Thanks!

PR #88, opened 2016-02-16
Node Tagging:

zyoung51:
- Modify node model to have a tags array and add $addToSet and $pull native
  mongo commands to modify the array
- Create a tags model with SKU model based rules

### DIFF
--- a/lib/models/node.js
+++ b/lib/models/node.js
@@ -8,10 +8,11 @@ NodeModelFactory.$provide = 'Models.Node';
 NodeModelFactory.$inject = [
     'Model',
     'Services.Waterline',
-    '_'
+    '_',
+    'Promise'
 ];
 
-function NodeModelFactory (Model, waterline, _) {
+function NodeModelFactory (Model, waterline, _, Promise) {
     return Model.extend({
         connection: 'mongo',
         identity: 'nodes',
@@ -54,6 +55,10 @@ function NodeModelFactory (Model, waterline, _) {
                 type: 'array',
                 defaultsTo: []
             },
+            tags: {
+                type: 'array',
+                defaultsTo: []
+            },
             // We only count a node as having been discovered if
             // a node document exists AND it has any catalogs
             // associated with it
@@ -64,6 +69,33 @@ function NodeModelFactory (Model, waterline, _) {
                     return !_.isEmpty(catalog);
                 });
             }
+        },
+        addTags: function(id, tags) {
+            // TODO: Refactor to native mongo command once it is merged
+            return Promise.fromNode(this.native.bind(this))
+                .then(function(collection) {
+                    return collection.update(
+                        { _id: waterline.nodes.mongo.objectId(id) },
+                        { 
+                            $addToSet: { tags: { $each: tags } },
+                            $set: { updatedAt: new Date() }
+                        });
+                });
+        },
+        remTags: function(id, tag) {
+            // TODO: Refactor to native mongo command once it is merged
+            return Promise.fromNode(this.native.bind(this))
+                .then(function(collection) {
+                    return collection.update(
+                        { _id: waterline.nodes.mongo.objectId(id) },
+                        { 
+                            $pull: { tags: tag },
+                            $set: { updatedAt: new Date() }
+                        });
+                });
+        },
+        findByTag: function(tag) {
+            return waterline.nodes.find({tags: tag});
         }
     });
 }
--- a/lib/models/tags.js
+++ b/lib/models/tags.js
@@ -1,0 +1,46 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = TagModelFactory;
+
+TagModelFactory.$provide = 'Models.Tag';
+TagModelFactory.$inject = [
+    'Model',
+    '_',
+    'Assert',
+    'Validatable',
+    'anchor'
+];
+
+function TagModelFactory (Model, _, assert, Validatable, Anchor) {
+    var allRules = _.keys(new Anchor().rules);
+
+    return Model.extend({
+        types: {
+            tagRules: function(rules) {
+                assert.arrayOfObject(rules, 'rules');
+                _.forEach(rules, function (rule) {
+                    assert.string(rule.path, 'rule.path');
+                    _(rule).omit('path').keys().forEach(function (key) {
+                        assert.isIn(key, allRules, 'rule.' + key);
+                    }).value();
+                });
+                return true;
+            },
+        },
+        connection: 'mongo',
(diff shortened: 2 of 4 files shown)


zyoung51:
@RackHD/corecommitters @uppalk1 @keedya @tannoa2

jlongever:
test this please

zyoung51:
test this please

spec/lib/models/node-spec.js
```diff
@@ -152,5 +152,72 @@ describe('Models.Node', function () {
...
+            });
+        });
+
+        describe('Node Tag Functions', function() {
+            var collection = {
+                update: sinon.stub().resolves()
+            };
+
+            before(function() {
+                sinon.stub(this.model, "native", function(cb) { cb(null, collection); });
```
heckj: `teachme` question - why stub out `model.native` like this? Not cluing in how it's used by the tests.... (not clear on what `model.native` is - are we just using it to make sure the collections got called and invoked?)

zyoung51: It is stubbed out with this pattern so the underlying code can promisify the function.  It is being stubbed out to remove the mongodb dependency from the unittest. Functional/integration testing would be used to validate MongoDB behavior/interactions, not unittests.

lib/models/tags.js
```diff
@@ -0,0 +1,47 @@
...
+TagModelFactory.$inject = [
+    'Model',
+    '_',
+    'Assert',
+    'Validatable',
+    'anchor',
+    'Promise'
+];
+
+function TagModelFactory (Model, _, assert, Validatable, Anchor, Promise) {
```
heckj: planning on using Promise later in here, or was this a "just in case" that didn't get used?

spec/lib/models/node-spec.js
heckj: No qualms with the tactics, just trying to grok where `.native` normally resides. I'm not clear on where it's getting promisified. Is native something that's already in the waterline model objects and is a convenient hook for capturing the flow that would otherwise go down into the MongoDB drivers?

lib/models/node.js
```diff
@@ -64,6 +69,31 @@ function NodeModelFactory (Model, waterline, _) {
                     return !_.isEmpty(catalog);
                 });
             }
+        },
+        addTags: function(id, tags) {
+            return Promise.fromNode(this.native.bind(this))
```
heckj: finally spotted the `native` here - guessing it'll be easier to catch you on chat to ask about this, but any pointers you can help me with are appreciated. is the `Promise.fromNode` what promisifyies the `this.native` as you're binding it?

lib/models/node.js
```diff
@@ -64,6 +69,31 @@ function NodeModelFactory (Model, waterline, _) {
...
+                .then(function(collection) {
+                    return collection.update(
+                        { _id: waterline.nodes.mongo.objectId(id) },
+                        { 
+                            $addToSet: { tags: { $each: tags } },
+                            $set: { updatedAt: new Date() }
+                        });
+                });
+        },
+        remTags: function(id, tag) {
```
anhou: remTags is names as 'removeTags' is more straightforward。

lib/models/node.js
```diff
@@ -64,6 +69,31 @@ function NodeModelFactory (Model, waterline, _) {
...
+            return Promise.fromNode(this.native.bind(this))
+                .then(function(collection) {
+                    return collection.update(
```
anhou: I'm not sure whether @benbp  have concerns about introducing mongodb native commands to RackHD compatibility in the future. 

it's ok for me, because waterline CRUD cannot cover the cases such as remove a key:value in one object here.
Just think that it's better to move mongodb native commands to some common place so that it could be leveraged by other codes, just like removing a 'key:value' here is a common case. and previous waterline CRUD message is published to amqp 
https://github.com/RackHD/on-core/blob/master/lib/common/model.js#L62-L89 , the native operations in a common place could also so similar things

lib/models/tags.js
zyoung51: I'll take it out.  I can add it back in later if/when I need it.

lib/models/node.js
zyoung51: Your opinion is appreciated, but rem/add is more straightforward to me.  
Bike-shedding: https://en.wikipedia.org/wiki/Law_of_triviality.

lib/models/node.js
zyoung51: @anhou: Ben did move the native commands to a common place in his workflow changes, but they are not in yet and I do not want to make my PR dependent on his.

lib/models/node.js
zyoung51: Promise.fromNode promisifies this: http://sailsjs.org/documentation/reference/waterline-orm/models/native.  Credit to @benbp for the work: https://github.com/RackHD/on-core/pull/75/files#diff-89e9e9d58a3c4aa21cf6a96f2bde6e15R110

lib/models/node.js
anhou: just the mind about code readability when first glance this line, at least for me, not for trivials, it's also fine for me if others also think it's fine

lib/models/node.js
anhou: just noticed @benbp 's PR's native commands codes. appreciate that, it makes sense that not depend that PR
:+1:

anhou:
:+1:

lib/models/node.js
benbp: Yeah, I don't really have a problem with remTags, it's just like "delX" which you see in a lot of places.

Also @zyoung51 if you get the chance will you add a // TODO above this to refactor it into this.updateMongo once my workflow PR goes in?

lib/models/node.js
benbp: @anhou yep, @tldavies is also going to do the same thing for the findAndModify atomic operation you suggested a while back for lookup entries on-dhcp-proxy I believe.

jlongever:
test this please

jlongever:
test this please

zyoung51:
test this please

jlongever:
test this please

jlongever:
@zyoung51 looks like on-taskgraph is failing unit-tests with on-core applied to it.  Would you mind checking it out locally?

zyoung51:
@jlongever It is passing on-taskgraph master w. on-core add-tagging locally.

jlongever:
test this please

zyoung51:
test this please

zyoung51:
test this please

jlongever:
test this please

jlongever:
+1

keedya:
:+1:

PR #89, opened 2016-02-24
Fetch all task definitions before doing cumulative updates to context…

benbp:
… object when validing properties

This fixes discovery on node v0.10. The context object is a shared object across multiple function calls when doing task graph property validation, but the ordering was dependent on the return order of a series of asynchronous calls, which was unreliable. This PR separates those two steps.

@RackHD/corecommitters @jlongever @johren @stuart-stanley @heckj @VulpesArtificem

### DIFF
--- a/lib/workflow/task-graph.js
+++ b/lib/workflow/task-graph.js
@@ -250,22 +250,32 @@ function taskGraphFactory(
             self._validateTaskLabels();
         })
         .then(function() {
+            assert.arrayOfObject(self.definition.tasks, 'Graph.tasks');
             return Promise.map(self.definition.tasks, function(taskData) {
-                if (_.has(taskData, 'taskDefinition')) {
+                if (!_.has(taskData, 'taskDefinition')) {
+                    return store.getTaskDefinition(taskData.taskName)
+                    .then(function(definition) {
+                        return {
+                            taskDefinition: definition,
+                            label: taskData.label
+                        };
+                    });
+                } else {
+                    return {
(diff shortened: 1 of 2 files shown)


jlongever:
+1

lib/workflow/task-graph.js
```diff
@@ -251,21 +251,36 @@ function taskGraphFactory(
...
+            .then(function(tasks) {
+                _.forEach(tasks, function(taskData) {
+                    if (_.has(taskData, 'taskDefinition')) {
+                        self._validateTaskDefinition(taskData.taskDefinition);
+                        self._validateProperties(taskData.taskDefinition, context);
+                        self._validateOptions(taskData.taskDefinition, taskData.label);
+                    } else {
+                        self._validateTaskDefinition(taskData.taskDefinition);
+                        self._validateProperties(taskData.taskDefinition, context);
+                        self._validateOptions(taskData.taskDefinition, taskData.label);
```
yyscamper: This `if/else` branches call the same functions with same arguments, so why need the branch?

heckj: they look identical to me too - @benbp did you mean to do something different with `taskData.taskDefinition` in the top stanza?

benbp: Whoops! There used to be a difference but now I normalize the data in that previous block, will update.

jlongever:
Function scripts pass with all dependencies locally.
+1 !

VulpesArtificem:
+1

PR #91, opened 2016-02-25
updated setIp in Lookup to now use findAndModifyMongo.

tldavies:


tldavies:
@benbp

tldavies:
@RackHD/corecommitters

lib/models/lookup.js
```diff
@@ -74,37 +75,45 @@ function LookupModelFactory (Model, assert, Errors, Promise, Constants) {
...
-                    return self.create({ ipAddress: ipAddress, macAddress: macAddress });
-                } else {
-                    return self.update({ id: macRecord.id }, { ipAddress: ipAddress });
+            };
+
+            var options = { new: true };
+
+            //Queries for the ipAddress that are not matched with the macAddress
+            //changes ip to null if macAddress doesn't match
+            waterline.lookups.findAndModifyMongo(query, {}, update, options);
```
benbp: This is async so you'll want to return it otherwise we can't catch errors or guarantee any ordering of this and the findAndModify query below.

benbp:
@anhou FYI

-1 because of the un-returned promises. Looks good otherwise.

benbp:
:+1:

lib/models/lookup.js
```diff
@@ -74,37 +75,46 @@ function LookupModelFactory (Model, assert, Errors, Promise, Constants) {
...
-                            })
-                    ];
-                }
-            }).spread(function(macRecord) {
-                if (!macRecord) {
-                    return self.create({ ipAddress: ipAddress, macAddress: macAddress });
-                } else {
-                    return self.update({ id: macRecord.id }, { ipAddress: ipAddress });
+            var query = {
+                ipAddress: ipAddress, // 10.1.1.2
```
anhou: the specific ip address'10.1.1.2' is confused. could you help refine the annotation?

lib/models/lookup.js
```diff
@@ -74,37 +75,46 @@ function LookupModelFactory (Model, assert, Errors, Promise, Constants) {
...
+                .then(function () {
+                    // update new document for new IP assignment, do this second
+                    query = {
+                        ipAddress: { $ne: ipAddress },
+                        macAddress: macAddress // new mac
+                    };
+
+                    update = {
+                        $set: {
+                            ipAddress: ipAddress // 10.1.1.2
```
anhou: the save with above about annotation

anhou:
:+1:

anhou:
test this please

lib/models/lookup.js
benbp: I think this comment isn't really needed at all actually.

anhou:
test this please

PR #94, opened 2016-03-02
add 'chai-string' to extend chai's string comparison assertions

yyscamper:
I see @srinia6 extends the Chai with [chai-string](http://chaijs.com/plugins/chai-string/) to help some string comparison assertions (see this PR: https://github.com/RackHD/on-http/pull/137). I think this is a good extension and will probably be used by other repos, so its better to move it to on-core. Meanwhile, keep the Chai plugin installation in the same place (together with `Chai-as-promise` and `Chai-Sion`) will be better.
@RackHD/corecommitters @srinia6 @iceiilin @WangWinson @cgx027 @pengz1 @sunnyqianzhang

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   "devDependencies": {
     "chai": "^2.0.0",
     "chai-as-promised": "^4.2.0",
+    "chai-string": "^1.2.0",
     "coveralls": "^2.11.4",
     "express": "^4.10.7",
     "grunt": "^0.4.5",
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -33,6 +33,7 @@ global.sinonPromise = require('sinon-as-promised')(bluebird);
 global.chai = require('chai');
 global.chai.use(require('chai-as-promised'));
 global.chai.use(require('sinon-chai'));
+global.chai.use(require('chai-string'));
 
 /**
  *  set up global expect for testing


benbp:
:+1:

jlongever:
test this please

tldavies:
It seems like Jenkins can't find chai-string

"Error: Cannot find module 'chai-string'"

jlongever:
The downstream repo applies on-core but isn't installing the chai-string module for some reason.

jlongever:
might need to clear out the npm cache,   test this please

jlongever:
Never mind the cache comment.  The problem is that the applied downstream project's modules are excluding on-core's devDependencies and chai-string is being referenced in helper.js.

-1 @yyscamper could you help resolve this?

jlongever:
@RackHD/corecommitters could I suggest that we don't add new packages unless there is an explicit use for them?

anhou:
@jlongever I also think so

yyscamper:
@jlongever , thanks for find the root cause, I just realized it's not easy to add a new and common module in unit-test, I have to add the new module in all downstream projects' package.json! 
So if all @RackHD/corecommitters don't disagree this change and do think it's worth to do it then I will take action to this chang to avoid idle work.

Actually, I think this is an enhancement opportunity and also a good way to reduce RackHD's footprint. Since all dev modules have been downloaded in on-core, so why we need another copy in each downstream project? I'll find a better way to handle this.

benbp:
@jlongever @yyscamper I was looking back at the change that introduced chai string in the first place, and it's just using `equalIgnoreCase` on a stringified array as a method to compare equality, which would be much better expressed as `.deep.equal`. So we don't really need chai-string or any of the updates at all, we just need to merge this fix: https://github.com/RackHD/on-http/pull/169

benbp:
Changing my :+1: to a :-1:, as we don't need chai string at all (see above comment).

yyscamper:
@benbp , OK we can defer the request until there is a case that really need this library. I'll close my PR.

PR #98, opened 2016-03-07
Fix lookup query atomicity and indexing. Update findAndModifyMongo da…

benbp:
…te modifiers

Resolves https://github.com/RackHD/RackHD/issues/121
Resolves https://github.com/RackHD/RackHD/issues/122

@RackHD/corecommitters @johren @heckj @tldavies

### DIFF
--- a/lib/common/model.js
+++ b/lib/common/model.js
@@ -129,6 +129,13 @@ function ModelFactory (
             return Promise.map(_.values(arguments), this.runNativeMongo.bind(this, 'createIndex'));
         },
 
+        createUniqueMongoIndexes: function() {
+            var self = this;
+            return Promise.map(_.values(arguments), function(arg) {
+                return self.runNativeMongo('createIndex', [arg, { unique: true }]);
+            });
+        },
+
         findMongo: function(query, options) {
             return this.runNativeMongo('find', [query])
             .then(function(cursor) {
--- a/lib/common/model.js
+++ b/lib/common/model.js
@@ -88,21 +88,22 @@ function ModelFactory (
             });
         },
 
-        addWaterlineDatesAsMongoModifiers: function(obj) {
-            obj.$set = obj.$set || {};
-            if (!obj.$set.updatedAt) {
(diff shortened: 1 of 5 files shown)


benbp:
The key changes here are an update to the queries in the lookup model, as well as adding a compound index for lookups. I also took the opportunity to improve waterline date fields based on various mongo queries.

DO NOT MERGE. I'm realizing I still need to enforce index uniqueness, going to make those changes now.

heckj:
@benbp do we have any unit tests or the like that can verify the schema validation out of this? (likely repeat now that you found the underlying code to tweak, but maybe useful?) and anything to validate the indices?

benbp:
@heckj I could add a test to the functional tests RackHD/RackHD/tests that just looks at mongo indexes after starting the services, but that doesn't guarantee that they weren't created elsewhere. Any further granularity would require something like what is being discussed at https://groups.google.com/forum/#!topic/rackhd/KuGYxlJFk-0.

The schema validation is also tricky for a similar reason. We need to verify it against what comes out of a real database, which can't be done in unit test. But I could add a functional test to make a lookup API call with a conflicting IP which would test the regression.

tldavies:
test this please

benbp:
CC @keedya @hohene

benbp:
@heckj digging into this, there's actually no good way to test this at all in the functional tests, because we don't expose this call currently in any testable part of the API.

heckj:
:+1:

anhou:
:+1:

yyscamper:
:+1:

jlongever:
Was this merged without functional tests passing?

zyoung51:
@jlongever on-http downstream unittests also fail with this change.  The waterline.lookups in start() is undefined.

zyoung51:
A fix for on-http is here: https://github.com/RackHD/on-http/pull/177

tldavies:
RackHD/on-http#177 has been merged to master

PR #100, opened 2016-03-10
As part of the swagger AuthN/AuthZ changes:

zyoung51:
- Add local user model and use bcrypt to serialize passwords
- Add roles model
- Add 401 and 403 errors to the common errors

### DIFF
--- a/lib/models/local-user.js
+++ b/lib/models/local-user.js
@@ -9,7 +9,19 @@ LocalUserModelFactory.$inject = [
     'Model'
 ];
 
-var bcrypt = require('bcrypt');
+var crypto = require('crypto');
+
+var hashConfig = {
+    // number of bytes in the Hash
+    hashBytes: 64,
+    // number of bytes in the salt
+    saltBytes: 64,
+    // number of iterations to get the final hash, longer means more secure,
+    // but also slower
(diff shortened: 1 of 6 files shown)


anhou:
:+1:

zyoung51:
test this please

lib/models/local-user.js
```diff
@@ -0,0 +1,78 @@
...
+                required: true,
+                minLength: 6,
+                maxLength: 50
+            },
+            role: {
+                type: 'string',
+                defaultsTo: 'ReadOnly'  // change to Constants
+            },
+            comparePassword: function (password) {
+                var combined = new Buffer(this.password, 'base64');
```
WangWinson: The compare password maybe a common function not limit to user model ? For example, we need to hash BMC password, too. So we may extract it as a common function later. This is nice to have.

WangWinson:
:+1:

lib/models/local-user.js
zyoung51: This compare password is using a one way hash function where-as the BMC passwords need to be support decryption so they can be used by the RackHD pollers, etc.

jlongever:
+1

PR #106, opened 2016-03-31
change redaction to support regex and expand the redaction list

yyscamper:
I see we forgot to add some sensitive words into the redaction list, for example:
 - "plainPassword" & "encryptedPassword" & "rootPassword" & "rootEncryptedPassword" for OS bootstrap jobs.
 - "serverPassword" for all Dell RACDAM tasks.
- "community" for snmp.

This will cause to leak those sensitive information as they will be recorded in log files.
If adding these words one by one, the redaction list will become so big, so I changed the redaction to support the regex.

@RackHD/corecommitters @iceiilin @cgx027 @sunnyqianzhang @pengz1

### DIFF
--- a/lib/common/constants.js
+++ b/lib/common/constants.js
@@ -39,7 +39,7 @@ function constantsFactory (path) {
                 Length: 6
             },
             Sanitizations: ['id', 'mac', 'ip'],
-            Redactions: [ 'password' ]
+            Redactions: [ /password/i, /^community$/i ]
         },
         Configuration: {
             Files: {
--- a/lib/serializables/log-event.js
+++ b/lib/serializables/log-event.js
@@ -139,16 +139,19 @@ function LogEventFactory (
 
     /**
      * Initialize the redact regex
-     * By initializing all the redact regex in advance can speed up the subsequent testing
+     *
+     * By initializing all the redact 'RegExp' pattern in advance can speed up the subsequent
+     * testing; Keep the 'String` pattern of its original type can leverage the string comparasion
+     * which is faster than regex.
      *
      * @param {Array} redactions - The array of regex
      */
     LogEvent.initRedact = function(redactions) {
-        LogEvent._redactRegexes = _.reduce(redactions, function(result, value) {
-            if (_.isRegExp(value)) {
+        LogEvent._redactPatterns = _.reduce(redactions, function(result, value) {
+            if (_.isString(value)) {
+                result.push(value); //exactly full word match
+            } else if (_.isRegExp(value)) {
                 result.push(new RegExp(value));
-            } else if (_.isString(value)) {
-                result.push(new RegExp('^' + value + '$')); //exactly full word match
             } else {
                 console.log("warning: unsupported redaction type '" + typeof(value) + "', " +
                     "only 'RegExp' and 'String' is allowed.");
@@ -166,8 +169,17 @@ function LogEventFactory (
         if (!_.isString(value)) {
             return false;
         }
-        return _.some(LogEvent._redactRegexes, function(reg) {
-            return reg.test(value);
+
+        return _.some(LogEvent._redactPatterns, function(pattern) {
+            if (_.isString(pattern)) {
+                // The reason why use 'valueOf':
+                // new String('abc') is not strictly equal to new String('abc'),
+                // but new String('abc').valueOf() is strictly equal to new String('abc').valueOf()
+                return (pattern.valueOf() === value.valueOf());
+            }
+            else { //Regex
+                return pattern.test(value);
+            }
         });
     };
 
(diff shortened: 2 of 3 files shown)


zyoung51:
Rather than redacting the information, should it just be omitted? Dumping out redacted values only accomplishes producing noisy output.  By definition of the act of redacting, we are producing no useful information in the output....

yyscamper:
@zyoung51 , My point is that using redaction can keep the information that the data is there but it is modified due to security concern, otherwise we cannot distinguish whether the user really input data or not.
For instance, the `rootPassword` is the required option for os-bootstrap, if we omit it in the log file, when the bootstrap fails, by examining the log, we may judge it is due to user doesn't input the `rootPassword`.

zyoung51:
@yyscamper Users should debug their inputs by examining the log file and not rely on our parameter checking? I'm not sure how your example is valid as required parameters are validated on input.

+0, this change adds noise to the log, reinforces the idea that our parameter checking is bad or absent, and compels a user to debug their input values by examining a log file rather than our api response body.  I still think the fields should be omitted and not logged at all.

yyscamper:
@zyoung51 I understand your concern, but personally I think the redaction is still better than totally omit it because, as you mentioned, the validation coverage is not so good right now, and even it has been ideally perfect(very difficult to become perfect), the redaction also provides double confirm.
 As for the noise you mentioned, it is very low, we don't frequently print those sensitive info and usually only few fields are redacted for one object printing.

I also invite other @RackHD/corecommitters to see their comment for this.

jlongever:
Hopefully nobody sets their password to "[REDACTED]".   I agree 100% with Zack's comments.  Let's fix the input validation rather than encouraging a user sift through debug logs to validate their input.

yyscamper:
@jlongever, @zyoung51, Indeed the input validation should be enhanced, but the redaction doesn't conflict with the input validation, how about if the `password` is optional, the input validation will let it pass, but if user inputs the password, we should still avoid logging its clear text.

If we choose to delete the sensitive field in log file rather than redaction, it will feel user confused if that object contains dozen properties but only the `password` is missing. Actually we had similliar redaction mechanism in on-http serializables:
- https://github.com/RackHD/on-http/blob/master/lib/serializables/v1/obm.js#L56
- https://github.com/RackHD/on-http/blob/master/lib/serializables/v1/snmp.js#L38

I'm OK if you don't like the word `[REDACTED]`, then how about `xxx` or `***`?

anhou:
IMHO, the validation of input and logs are two things. the input should be validated, and the logs for debug, analysis or  record, both are needed. if the 'password' key is one key of the printed object, if all other object's keys are printed except 'password' key/value, maybe confuse users or developers when they see there's no 'password' key. for what value it should be, I think [REDACTED], or 'xxx', or '***' , or any encrypted value are all fine. but maybe [REDACTED] is better to keep consistent with other parts of RackHD.
so :+1:  from me for this part.

jlongever:
@yyscamper @anhou FWIW,  if you're looking for some kind of convention here both the Redfish spec and OpenStack excludes the password field entirely;  I see no value in including it anywhere or introducing more code to maintain to handle redaction.

brianparry:
@yyscamper Personally, I think it's unlikely that displaying [REDACTED] will add meaningful debug information.  In fact, I'd say there is a problem if it is; if we need to depend on seeing [REDACTED] in the logs, we're missing proper debug hints elsewhere.  While I agree that the noise resulting from this change will be minimal, I think it's noise nonetheless.  So, +0 from me.

zyoung51:
+0 from me.  Ultimately, the code should be sanitizing and omitting the sensitive values in waterline with .toJSON() (or equivalent elsewhere) before logging the object.

In my opinion, this PR does not address the root cause issues:
- Incomplete parameter checking resulting in hard to debug issues that require examination of log files.
- Improper usage of logger by logging unsanitized objects

This PR adds a potential new issue:
- Increased logger load (performance reduction?) due to regex comparisons on all objects that are logged.

yyscamper:
Thanks for you further comment @jlongever @brianparry @zyoung51, I understand the concern that this redacted data may not be valuable if only for debugging purpose.  I just guess the intention when @jfrey wrote the first line of this redaction code, besides hiding sensitive data, this can also avoid to change those waterline toJSON one by one, also some object are not in database, such as the `options` that user passes to a workflow.

I also thought about the performance lost, previously the redaction is judged by comparing string rather than regex, the regex indeed have some performance lost, the benefit we get is the redaction words list can be vastly reduced. I plan to directly use the string comparison if user defines redaction words as string (currently I change the string to regex). This is a compromise between performance and usability.

yyscamper:
I made change to leverage the String comparison if the redaction keys are defined in String, so this will gain a little performance increase. 

Though most people don't like this redaction design but also no objection, I planed to first merge this PR to keep the redaction as it is before and in future create another PR to enhance this redaction design after we make agreement.

yyscamper:
Story has been created to track the follow-up task: https://www.pivotaltracker.com/projects/1492892/stories/117317989

PR #107, opened 2016-03-31
Add Constants.Task.ActiveStates

brianparry:
* Add new ActiveStates constant.
* Change graph object model active method to use this constant.

### DIFF
--- a/lib/common/constants.js
+++ b/lib/common/constants.js
@@ -171,6 +171,9 @@ function constantsFactory (path) {
                 'timeout',
                 'cancelled'
             ],
+            ActiveStates: [
+                'pending'
+            ],
             States: {
                 Running: 'running',
                 Succeeded: 'succeeded',
--- a/lib/models/graph-object.js
+++ b/lib/models/graph-object.js
@@ -41,7 +41,7 @@ function GraphModelFactory (Model, Constants) {
             },
             active: function() {
                 var obj = this.toObject();
-                return obj._status === Constants.Task.States.Pending;
+                return Constants.Task.ActiveStates.indexOf(obj._status) > -1;
             }
         }
     });


brianparry:
@RackHD/corecommitters @srinia6 @abildw @geoff-reid @dalebremner

heckj:
Setting up to add other states as "ActiveStates"?

brianparry:
@heckj Mainly I wanted to explicitly call out the set of active states for queries like this: https://github.com/brianparry/on-http/blob/workflows-rebased/lib/services/workflow-api-service.js#L241
I figured there was a reasonable possiblitity that we may want to add additional active states in the future (e.g. "running"), so I defined ActiveStates as an array.

heckj:
:+1:

zyoung51:
@brianparry fwiw, queries like that should probably talk to the workflow store rather than going to waterline directly.  For example, performing an IN operation like that is only valid on primary keys in Cassandra so additional logic would be needed to make that query functional for that type of backend.

zyoung51:
+1

brianparry:
test this please

benbp:
Ah I see what you were saying in that other PR now. :+1:, and as you can probably predict I agree with Zack's comment.

brianparry:
@zyoung51 @benbp I agree in principle, but there were a few complications to doing this.  In fact, the referenced query was originally done in the Mongo store, but for reasons mentioned here https://github.com/RackHD/on-http/pull/194, we opted to use a waterline query rather than `findOneMongo`.  Since the store uses a native Mongo query, we had the options of changing the store to use waterline, patching up the Mongo document to look like a waterline object, adding another similar function to the store that queried waterline rather than Mongo, or calling directly into waterline.  The latter option seemed simplest and least likely to have unexpected side effects.

benbp:
@brianparry just saw the updated query, +1 since it removes the performance concerns I had (offloading the work to the db rather than iterating over an unbounded array of objects). It still might cause us problems in regards to database flexibility (since the query isn't hidden behind an interface), but that's something we can tackle later.

brianparry:
Sounds good @benbp.  Thanks!

PR #112, opened 2016-04-13
Additional error handling for rendering views

brianparry:
* Add ViewRenderError and InternalServerError with 500 status.
* Add check for valid template in db renderable render method.

### DIFF
--- a/lib/common/db-renderable-content.js
+++ b/lib/common/db-renderable-content.js
@@ -12,7 +12,8 @@ dbRenderableContentServiceFactory.$inject = [
     '_',
     'Services.Waterline',
     'Logger',
-    'ejs'
+    'ejs',
+    'Assert'
 ];
 
 /**
@@ -33,7 +34,8 @@ function dbRenderableContentServiceFactory(
     _, 
     waterline, 
     Logger, 
-    ejs ) 
+    ejs,
+    assert )
 {
     var logger = Logger.initialize(dbRenderableContentServiceFactory);
 
@@ -118,6 +120,7 @@ function dbRenderableContentServiceFactory(
 
     DbRenderableContentService.prototype.render = function render (name, options, scope) {
         return this.get(name, scope).then(function (template) {
+            assert.ok(template, name + ' is not a valid template');
             return ejs.render(template.contents, options);
         });
     };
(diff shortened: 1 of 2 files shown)


lib/common/db-renderable-content.js
```diff
@@ -118,6 +118,11 @@ function dbRenderableContentServiceFactory(
 
     DbRenderableContentService.prototype.render = function render (name, options, scope) {
         return this.get(name, scope).then(function (template) {
+            if (!template) {
+                throw new Errors.TemplateRenderError(
```
heckj: need to inject an Error object in the `di` setup to get to this?

brianparry: @heckj I was hoping no one would notice that :)  I wasn't quite quick enough pushing the fix.

heckj: lol, all good

lib/common/db-renderable-content.js
```diff
@@ -118,6 +120,11 @@ function dbRenderableContentServiceFactory(
 
     DbRenderableContentService.prototype.render = function render (name, options, scope) {
         return this.get(name, scope).then(function (template) {
+            if (!template) {
```
benbp: Maybe this is way too minuscule of a concern, but wouldn't we want empty strings to be valid templates?

brianparry: @benbp Yeah - I guess that's true; an empty string is a valid template.  I added this check to catch the case where template is undefined.  Maybe `typeof(template) === 'string'` is a more appropriate check.

brianparry: @benbp I take it back.  `template` is an object; the renderable string is referenced by the `contents` property, so an empty string will still pass this test.

zyoung51: I called the variable 'template', but it can be a template, profile or view.  Should we be throwing a TemplateRenderError when it can be either of the three?  Maybe we just want to use asserts and then let the parent object handle throwing the more specific error (if it is needed) or use the collectionName to throw the most correct collection specific error?

brianparry:
test this please

brianparry:
test this please

yyscamper:
:+1:

brianparry:
test this please

srinia6:
+1

PR #113, opened 2016-04-14
Add 409 conflict error to Errors

zyoung51:


heckj:
👍

brianparry:
:+1:

benbp:
test this please

yyscamper:
:+1: after adding the unit-test to improve the coverage.

I see there are multiple PRs to add new Error recently, I have some proposals to improve this Errors definition (not required for this PR), and I want to hear your voice of @RackHD/corecommitters and @zyoung51 :
(1) Separation the HTTP erros and other erros, create HttpBaseError (all Http Error has a `status` prooperty), HttpBaseError inherits from BaseError, and all other Http Errors inherits from HttpBaseError. 
(2) Create all HTTP normal errors at a time, not one by one, thus minimize the times of such across repos changes in future.
(3) Improve the Error unit-testing, make it be more flexible to add new Error type.

zyoung51:
@yyscamper What do you suggest is unittested here?  The fact that status is a 409 value?

@RackHD/corecommitters Closing the PR.  I'll use BaseError and set a status property to 409 instead.

amymullins:
@yyscamper and @heckj  can you help provide some standardized criteria on percentage of unit test (Coveralls) coverage loss that should gate a PR?  This PR combined with (https://github.com/RackHD/on-core/pull/112/files) had a -0.09% loss of unit test coverage.  This PR was flagged while the other was given +1s and not gated.  

Also, @yyscamper for your refactoring suggestion would you mind to put that in google groups?  Those comments will be lost if/when this PR is merged.

zyoung51:
@RackHD/corecommitters: Updated on-http dependent PR to indicate that it would be changed after Felix has refactored the error handling: '// TODO: Replace with HttpError after errors are refactored' and switched to using BaseError and setting status to 409.

yyscamper:
@amymullins , @zyoung51 : Actually we don't have a common criteria on percentage of unit test coverage, my practice is that it differs in following situation: 
(1) If the new code totally doesn't have any unit-test code, then it is gating; 
(2) If there is a lot of new code and reasonable amount of unit-test, then it is OK for me as long as the coverage doesn't decrease a lot ( less than around 0.5% in my practice). 

Anyway, the more the unit-test, the better for our project. And I just found a bug (see https://github.com/RackHD/RackHD/issues/186), if we had written the unit-test, that bug can be prevented.

As for those suggestion about Error refactor, I will write them in google group, so we don't forget it.

amymullins:
Thanks for your guidelines @yyscamper !  In this case, I believe the coverage loss with the combined 2 PRs was 0.09% and was considered negligible test coverage loss by both developers.  If applying your guidelines, then going forward we may not want to waste valuable CC time with such a comment given the large amount of PRs these days, let the automatic build tools call out the coverage loss, and review the content of the unit tests to make sure there is sufficient coverage.  

I also agree 100% on adding the supporting unit test coverage, that is part of our definition of done and should have technically gated BOTH PRs.  Glad it was caught in code review for this PR, but was missed in its peer: https://github.com/RackHD/on-core/pull/112/files.

PR #114, opened 2016-04-14
Removes event loop blocked detection.

rolandpoulter:
@RackHD/corecommitters

From a conversation with @benbp about changing the threshold for "Event Loop Blocked" appearing in the logs. Ultimately I thought we should remove this, because I believe it really does more harm than good. Mainly because `setInterval` is not very accurate, and even with an idle RackHD you'll still see "Event Loop Blocked" (50ms) debug logs.

I've seen people in RackHD ask about these logs, and that makes me feel they are somewhat confusing and noisy. Not really something that should be running in production.

We should use data from our automated tests to figure out where bottlenecks are and not rely on this kind of mechanism.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,6 @@ module.exports = function (di, directory) {
             helper.requireWrapper('apache-crypt', 'apache-crypt'),
             helper.requireWrapper('util', 'util'),
             helper.requireWrapper('pluralize', 'pluralize'),
-            helper.requireWrapper('blocked', 'blocked'),
             helper.requireWrapper('always-tail', 'Tail'),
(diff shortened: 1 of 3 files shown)


heckj:
I'm good with this, but let's put up a relevant PR to remove the explanation for this as well - referenced docs section:
 - https://github.com/RackHD/docs/edit/master/docs/devguide/debugging_guide.rst - line 126 to 153

benbp:
👍

yyscamper:
:+1:

yyscamper:
test this please

brianparry:
test this please

PR #118, opened 2016-04-21
Set recursive to true when load templates

WangWinson:
The `FileLoader` [getAll] (https://github.com/RackHD/on-core/blob/master/lib/common/file-loader.js#L59) method has already support load file recursively in sub folder, set `recursive` true to enable add templates and profiles in sub folder.

@RackHD/corecommitters @sunnyqianzhang @iceiilin @cgx027 @pengz1


sunnyqianzhang:
+1. Verified and passed with Ubuntu installation on Hydra platform

benbp:
👍

PR #122, opened 2016-04-26
export InternalServerError from Errors injectable (1.2.0)

brianparry:


brianparry:
test this please

yyscamper:
:+1: after adding the unit-test to avoid the previous Error type is not defined problem.

brianparry:
@yyscamper I'd suggest adding unit tests for errors as part of the error refactoring you suggested in google groups.

PR #125, opened 2016-04-28
Add module to read local system unique identifier

jlongever:
@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @zyoung51 @geoff-reid 

This PR is required for upcoming Redfish ServiceRoot fix and SSDP/UPnP PR.

### DIFF
--- a/lib/common/system-uuid.js
+++ b/lib/common/system-uuid.js
@@ -9,10 +9,11 @@ SystemUuidFactory.$inject = [
     'uuid',
     'fs',
     'Logger',
-    'Promise'
+    'Promise',
+    'Assert'
 ];
 
-function SystemUuidFactory(uuid, nodeFs, Logger, Promise) {
+function SystemUuidFactory(uuid, nodeFs, Logger, Promise, assert) {
     var logger = Logger.initialize(SystemUuidFactory);
     var fs = Promise.promisifyAll(nodeFs);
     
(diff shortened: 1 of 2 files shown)


lib/common/system-uuid.js
```diff
@@ -0,0 +1,39 @@
...
+        // Attempt to get the system's UUID from sysfs otherwise generate one
+        return fs.readFileAsync('/sys/class/dmi/id/product_uuid')
+        .then(function(content) {
+            if(content) {
+                return content.toString().toLowerCase().replace(/(\r\n|\n|\r)/gm,'');
+            }
+            logger.warning('Error getting system UUID, generating...');
+            return uuid('v4');             
+        })
+        .then(function(uuid) {
```
zyoung51: Do you need a .catch() to handle file read errors like EACCES?

spec/lib/common/system-uuid-spec.js
```diff
@@ -0,0 +1,43 @@
+// Copyright 2016, EMC, Inc.
+'use strict';
+
+describe("SystemUuid", function() {
+    var subject;
+    var regEx = /^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$/;
```
yyscamper: This regex is not the same with https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L197

yyscamper:
:+1: after change the regex to the common one.

spec/lib/common/system-uuid-spec.js
jlongever: It is the same as assert-plus's assert.uuid().   https://github.com/mcavage/node-assert-plus/blob/master/assert.js#L12

Are you suggesting we're not asserting UUID values correctly and shouldn't be using assert.uuid()?

jlongever:
test this please

spec/lib/common/system-uuid-spec.js
brianparry: @jlongever, @dalebremner added a regex for UUID to Constants.Regex.  I think Felix is suggesting you use that one.  At first glace, they appear the same.

PR #127, opened 2016-04-29
Changes to work-item model

brianparry:
This patch includes changes to the work-item model that are needed to remove serializables from the 2.0 API pollers route. Changes convert the pollers resource "type" property to "name" and redact passwords from poller config.


brianparry:
This test failure is caused by password redaction.  It will need to be coordinated with a test change.

brianparry:
Depends on https://github.com/RackHD/RackHD/pull/208.  This will fix the functional test failure.

srinia6:
+1

PR #128, opened 2016-04-29
Bump mocha timeout to 10000 to mitigate potential jenkins issues

benbp:
Proposing a potential way to address what could be jenkins timeout issues when loading modules in mocha "before" blocks.

@RackHD/corecommitters @tldavies @davidzuhn

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "doc": "jsdoc lib -r -d docs",
-    "test": "./node_modules/.bin/mocha $(find spec -name '*-spec.js') -R spec --require spec/helper.js"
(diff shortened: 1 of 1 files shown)


tldavies:
👎  This should try to be fixed via jenkins with lowering the node count for the build slave first.

davidzuhn:
What do you mean by "lowering the node count for the build slave"?

tldavies:
So the old build slave for 1.1 release only have 2 nodes. While the new build slave has 4. I would lower that count to 2 again. As that worked in the passed when we had timeouts in jenkins.

davidzuhn:
Do you mean cpus?

tldavies:
screen shot 2016-05-02 at 3 32 03 pm
The above photo shows 4 idle reduces it to 2.

davidzuhn:
Most of the build failures have occurred when that build job was the only one running on the given node.

PR #131, opened 2016-05-09
Add support for southbound API proxy server

shanemsully:
In order to support running the RackHD on-taskgraph and on-http services on a host that does not have Layer-2 network access to the PXE network we need a host that does have access to the PXE network and can proxy requests to the southbound API endpoint over a management network.  

The proxy server, nginx in my POC, adds headers to the requests coming from the PXE client that contain headers that will be used during profile, template, and task rendering.

The headers are
X-Real-IP - the actual IP address of the PXE client
X-RackHD-api-proxy-ip - The IP address of the API proxy server
X-RackHD-api-proxy-port - The port of the API proxy server

The proxy server URL is added to the lookup record for the PXE client so that it can be used during task rendering since the original HTTP request is not available at that time.

on-http PR: https://github.com/RackHD/on-http/pull/274
on-tasks PR: https://github.com/RackHD/on-tasks/pull/185

This is one of three pull requests.  There will also be subsequent changes to on-http and on-tasks to support this.
[test-coverage.tar.gz](https://github.com/RackHD/on-core/files/255346/test-coverage.tar.gz)

### DIFF
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -48,6 +48,10 @@ function lookupServiceFactory(
     function remoteAddress(req) {
         assert.ok(req, 'req');
 
+        if(req.get("X-Real-IP")) {
+            return req.get("X-Real-IP");
+        }
+        
         if (req.ip) {
             return req.ip;
         }
@@ -360,6 +364,23 @@ function lookupServiceFactory(
         });
     };
 
+    /**
+     * Returns the proxy server for the given Node ID.
+     * @param  {String} id Node ID
+     * @return {Promise.<String>} A promise fulfilled to the Proxy server URL
+     * that corresponds to this node
+     */
+    LookupService.prototype.nodeIdToProxy = function (id) {
+
+        return waterline.lookups.findOneByTerm(id).then(function (record) {
+            if (record && record.proxy) {
+                return record.proxy;
+            } else {
+                return undefined;
+            }
+        });
+    };
+    
     LookupService.prototype.setIpAddress = function (ip, mac) {
         return waterline.lookups.setIp(ip, mac);
     };
--- a/spec/lib/services/lookup-spec.js
+++ b/spec/lib/services/lookup-spec.js
@@ -339,7 +339,12 @@ describe('Lookup Service', function () {
 
             this.sandbox.stub(lookupService, 'ipAddressToMacAddress').resolves('00:11:22:33:44:55');
 
-            var req = { ip: '10.1.1.1' },
+            var req = { 
+                    ip: '10.1.1.1',
+                    get: function() {
+                        return undefined;
+                    } 
+                },
                 next = function () {
                     expect(req.macaddress).to.equal('00:11:22:33:44:55');
                     expect(req.macAddress).to.equal('00:11:22:33:44:55');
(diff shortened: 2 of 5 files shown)


lib/services/lookup.js
```diff
@@ -360,6 +364,24 @@ function lookupServiceFactory(
...
     };
 
+    /**
+     * Returns the proxy server for the given Node ID.
+     * @param  {String} id Node ID
+     * @return {Promise.<String>} A promise fulfilled to the Proxy server URL
+     * that corresponds to this node
+     */
+    LookupService.prototype.nodeIdToProxy = function (id) {
+        assert.isMongoId(id);
```
benbp: We're trying to remove references to `isMongoId` as we start to add support for other databases.

lib/services/lookup.js
```diff
@@ -360,6 +364,24 @@ function lookupServiceFactory(
...
+
+        return waterline.lookups.findOneByTerm(id).then(function (record) {
+            if (record.proxy) {
```
benbp: Might be worth doing `if (record && record.proxy)` just to be safer.

spec/lib/services/lookup-spec.js
```diff
@@ -369,7 +379,34 @@ describe('Lookup Service', function () {
...
 
-            var req = { connection: { remoteAddress: '10.1.1.1' } },
+            var req = { 
+                    connection: { remoteAddress: '10.1.1.1' },
+                    get: function() {
+                        return undefined;
+                    }
+                },
+                next = function () {
+                    expect(req.macaddress).to.equal('00:11:22:33:44:55');
```
benbp: When using `done()` I usually wrap these in a try/catch, otherwise on failure cases the test just times out rather than showing the real assertion error, e.g.:

```
next = function () {
    try {
        <expect>
        done();
    catch (e) {
        done(e);
    }
};
```

benbp:
This seems like a pretty useful feature to have in the system. Thanks for taking the time to add the functionality! I've got a couple minor comments across this and the other two PRs. It might also be worth adding a documentation note somewhere in https://github.com/RackHD/docs that seems appropriate.

benbp:
FYI @brianparry @zyoung51 @rolandpoulter may be of interest.

jlongever:
Would be using something more conventional like HTTP X-Forward-For (https://en.m.wikipedia.org/wiki/X-Forwarded-For) be better rather than adding custom header fields?

zyoung51:
I agree with Joel on the usage of X-Forward-For and Nginx's X-Real-IP (Apache's X-Client-IP?) over RackHD custom naming for the standard functions.  @shanemsully Can you rebase this to master and resolve the merge conflicts?

shanemsully:
I did consider using the X-Forwarded-For header.  There were a few reasons that I went with custom headers.

1. To the standard the header will contain the client IP that the request was forwarded for it will no contain the IP address of the proxy server that forwarded the request.
2. If the request goes through a chain of proxies any one of them could add an X-Forwarded-For header that spoofs the client IP address.
3. Using custom headers enables the user to tell the clients to use a different API endpoint than the proxy that forwarded the request.

Anyways, I could just customize the NGINX (or whatever load balancer) config to set a header like:

X-Forwarded-For: <real client IP>,<API proxy IP>

Since we need some custom configuration for this in either case it seemed preferable to just use custom header names.

spec/lib/services/lookup-spec.js
yyscamper: Got a new skill :)

shanemsully:
@zyoung51 @jlongever @benbp Do you agree with my reasoning above for going with custom headers as opposed to X-Forwarded-for?  Anything else that is needed to merge this PR?

zyoung51:
@shanemsully Thanks for the detailed explanation.  PR contents look good to me.
+1

brianparry:
ok to test

jlongever:
test this please

brianparry:
Jenkins, test this please

yyscamper:
:+1:

brianparry:
The last test failure was unrelated to this change and is fixed by this PR: https://github.com/RackHD/RackHD/pull/245.  So Jenkins, test this please.

PR #132, opened 2016-05-09
added unique flag for indexes

VulpesArtificem:
changed the mongo store setIndexes function to read a 'unique: true' flag and call the underlying method for  creating unique indexes in order to restore uniqueness constraints lost in the Big Taskgraph Refactor.
supports https://github.com/RackHD/on-taskgraph/pull/96
@RackHD/corecommitters @rolandpoulter

### DIFF
--- a/lib/workflow/stores/mongo.js
+++ b/lib/workflow/stores/mongo.js
@@ -51,27 +51,13 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
     /**
      * @param {Object} indexObject an object with keys that correspond to the mogo
      * collections on which to place indexes and values that are arrays of indexes
-     * @example {
-     *       taskdependencies: [
-     *           {index: {taskId: 1}, unique: true},
-     *           {index: {graphId: 1}},
-     *           {index: {taskId: 1, graphId: 1}}
-     *       ],
-     *       graphobjects: [
-     *           {index: {instanceId: 1}, unique: true}
-     *       ]
-     *  }
      * @memberOf store
      * @returns {Promise}
      */
     exports.setIndexes = function(indexObject) {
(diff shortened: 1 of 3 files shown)


lib/workflow/stores/mongo.js
```diff
@@ -52,22 +52,26 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
...
-     *          taskdependencies: [
-     *              {taskId: 1},
-     *              {graphId: 1},
-     *              {taskId: 1, graphId: 1}
-     *          ],
-     *          graphobjects: [
-     *              {instanceId: 1}
-     *          ]
+     *       taskdependencies: [
+     *           {index: {taskId: 1}, unique: true},
```
zyoung51: Would it make more sense to keep this type of information in the model?  Is this really a mongo only behavior?

If all databases will need taskId to be unique, then the waterline 'unique' attribute could be set to 'true' in the model.  If all databases should index this key/row, then the waterline 'index' attribute could be set to 'true'.

benbp: An outstanding question might be does our mongo pass-through mean we still automatically create these indexes if they are in a waterline model, and also is there a way to make compound indexes via waterline, or does that stay with the mongo-specific implementation. If it all works with waterline, great, if not, perhaps we might want to at least implement our own abstraction at the model layer?

VulpesArtificem: Ok, there was a bit of misunderstanding on my part as to the actual bug here. Yes, we do create indexes/enforce uniqueness if we add them to the waterline model, however waterline does not support compound indexes. I've edited to use the waterline models for all simple index/unique necessities and use the mongo method to set the one compound index we want.

VulpesArtificem:
test this please

anhou:
:+1:

VulpesArtificem:
test this please

PR #133, opened 2016-05-11
this updates always-tail to the newest version 0.2.0. 

tldavies:
The new version does not throw a String when it cant find the file to tail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   },
   "homepage": "https://github.com/RackHD/on-core",
   "dependencies": {
-    "always-tail": "^0.1.1",
+    "always-tail": "^0.2.0",
     "amqp": "0.2.3",
     "anchor": "^0.10.2",
     "apache-crypt": "1.0.9",


heckj:
@tldavies I'm presuming this is resolving a bug or something - reference? Or is this technical debt work?

(trying to understand if there should be a matching PR for this against the 1.2 branch)

zyoung51:
Can this dependency be moved to on-dhcp-proxy along with the dhcp polling change?

tldavies:
@heckj The problem is that if you are testing or debugging objects that use this as a dependency it throws a double error. It throws a String and then the actual error. The following is what you get:

"Error: the string "Filename is missing: true" was thrown, throw an Error :)"

A fixed for this was merged in Sept 2014 and is included in the newest version that was published 7 months ago. 

@zyoung51 I'm ok with that as long as nothing else is using. I'll check to make sure.

tldavies:
@zyoung51 I could not find this used in anything besides for the isc-dhcp-poller. So I ok with removing it from on-core and move it to on-dhcp-proxy once the PRs for isc-dhcp-poller move have been merged.

tldavies:
@zyoung51 since we aren't moving the isc-dhcp-poller we should update this in on-core for now. Or move it to on-tasks.

anhou:
at present, isc poller is disabled by default. https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/isc-dhcp-poller-service-graph.js#L8   and it's a minor change, so it won't affect 1.2 release.  and I think it doesn't need to change more to migrate the wrapper iin on-core to on-tasks for unused codes  so :+1:  from me. or close it.

jlongever:
test this please

yyscamper:
It's +0 if we plan to get rid of isc-dhcp-poller (because we already have arp poller)
It's +1 if we want to keep isc-dhcp-poller. 

@jlongever : could you please confirm the fate of the isc-dhcp-poller? Personally I think we don't need to retain it.

jlongever:
@yyscamper I don't think DHCP polling can be completely removed as some network configurations may require it. For example https://github.com/RackHD/RackHD/issues/225

@tldavies had moved DHCP polling into on-dhcp-proxy with a config option and this PR adds a config option for using ARP cache: https://github.com/RackHD/on-core/pull/145.   @benbp?

heckj:
@yyscamper see the bug that Joel referenced - the dhcp poller is still desired/needed by folks using RackHD...

tldavies:
@jlongever - We closed the PR to move the on-dhcp-proxy with a config option, until ARP-Poller is moved. Will reopen since it is been moved. :-)

tldavies:
test this please

jlongever:
@tldavies +1

yyscamper:
:+1: 
@tldavies : please resolve the conflict.

tldavies:
@RackHD/corecommitters when this passes, please merge

tldavies:
This is failing cause of the md5 Hash

"AssertionError: expected '$1$WeJknBPk$Bjnh76GiuzCAoq4Gp6y.V0' to equal '$1$WeJknBPk$t3LdooA1UatyI0C7pGX5F/'
    at Context.&lt;anonymous&gt; (spec/lib/common/encryption-spec.js:93:29)"

tldavies:
test this please

benbp:
test this please

benbp:
test this please

anhou:
:+1:

PR #139, opened 2016-05-13
Removing tests for md5 as it is not longer supported apache-crypto pa…

tldavies:
…ckage

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "always-tail": "^0.1.1",
     "amqp": "^0.2.3",
     "anchor": "^0.10.2",
-    "apache-crypt": "^1.0.9",
+    "apache-crypt": "1.0.9",
     "assert-plus": "^1.0.0",
     "bluebird": "^2.8.0",
     "colors": "^1.0.3",


tldavies:
@rolandpoulter @VulpesArtificem @jlongever 
@RackHD/corecommitters

VulpesArtificem:
👍

rolandpoulter:
:+1:

tldavies:
^timed on the smoked tests.

tldavies:
test this please

zyoung51:
Test this please

jlongever:
test this please

jlongever:
+1

yyscamper:
@tldavies : I think following code and comment need be updated as well:
- https://github.com/RackHD/on-core/blob/master/lib%2Fcommon%2Fencryption.js#L95
- https://github.com/RackHD/on-core/blob/master/lib%2Fcommon%2Fencryption.js#L75
- https://github.com/RackHD/on-core/blob/master/lib%2Fcommon%2Fencryption.js#L87

:+1: after above code is fixed as well.

jlongever:
What is the value of testing the underlining apache-crypto library code?  These should probably just be mocked out, avoiding future failures due to unexpected changes--better yet,  lock to a specific version and opt-in when a new version is released.

zyoung51:
Agree with @jlongever, but also wonder what is the value of apache-crypt at all?  It is used on this one line: https://github.com/RackHD/on-core/blob/fea46c9e5c7b3f665abe47cd6ec68f27a0c8b3cc/lib/common/encryption.js#L81 and could be replaced with the crypto functions in nodejs (elsewhere we just use pbkdf2 for a one-way hash).  What is the value in using the library over the already available functionality in node?

jlongever:
@zyoung51 good point.

yyscamper:
@zyoung51 : Actually I'm  the first person which wrote the first code about creatHash and choosed the external library.

I had struggled to replace the external crypto library with nodejs builtin, but failed. Now  the createHash is created by install-os to encrypt the password before storing it into the kickstart. I don't quite understand the crypto algorithm, though they both called "sha512", but I cannot get a way to generate the Hash string (using same salt) that the installer can recognize using builtin crypto. If you had insight on this, I will be very appreciate.

tldavies:
@yyscamper Removed from the other files too. Thanks!

@jlongever @zyoung51 The tests is supposed to be testing all the options of createHash. I would say just remove the following, as you are right it only tests the crypto library.

https://github.com/RackHD/on-core/blob/master/spec/lib/common/encryption-spec.js#L86

That being said. We should merge this fix now, as it is breaking on-core and then remove/change how we test this class.

yyscamper:
@jlongever : Ahh, I'm also the first person who wrote those unit-test.
I think these unit-test is useful and I recommend to keep them. The reason is:
The external library has been changed multi time, the first is crypt3, but then we found it doesn't well on node4, then rolander trying to make a patch to fix it (see https://github.com/RackHD/on-core/pull/12), and finally we switch to apache-crypto. Meanwhile, we are still hoping to using the built-in crypto in a future day.. So the unit-testing is just trying to the functionality doesn't be broken even we changed the library.

yyscamper:
@tldavies I recommend to keep these unit-test, I just replied the reason to @jlongever in last comment.

tldavies:
@RackHD/corecommitters merge - to fix on-core

lib/models/file.js
```diff
@@ -31,11 +31,6 @@ function FileModelFactory (Model) {
                     uuidv4: true
                 },
 
-                md5: {
```
zyoung51: This should stay in the model.  on-http is updating the md5sum using crypto (https://github.com/RackHD/on-http/blob/fea2324006a440f2ed8f48972502698f34983de6/lib/services/files/file-plugin.js#L146) and we reference the md5sum in the workflows.

spec/lib/models/file-spec.js
```diff
@@ -77,21 +77,6 @@ describe('Models.File', function () {
             });
         });
 
-        describe('md5', function() {
```
zyoung51: A patch is needed to remove the md5 references from on-http and on-skupack and an accompanying patch to add sha256 computation to the microkernel before this can be removed.

tldavies:
@zyoung51 ok re-added them. Until we also remove them out of on-skupack and on-http.

zyoung51:
test this please

tldavies:
@yyscamper @zyoung51 @jlongever @benbp So here are the 2 options to get master back to passing:

1. hardcode to apache-crypt 1.0.9
2. Merge this PR and remove all other md5 references. 

Either way on-core needs to be fixed.

tldavies:
For now we are going to lock the version at 1.0.9 before md5 was taken out to fix on-core.

tldavies:
test this please

tldavies:
test this please

yyscamper:
@tldavies , I agree to lock the version at 1.0.9 because other modules are still using the md5.

zyoung51:
Test this please

PR #140, opened 2016-05-17
Fix graph created event, and add label to task instances.

rolandpoulter:
* Adds task labels to task instance data. (required by the UI for operation center)
* Re-implements created waterline event when graphs are created AND updated. (required by the UI for realtime graphs)

### DIFF
--- a/lib/workflow/stores/mongo.js
+++ b/lib/workflow/stores/mongo.js
@@ -15,6 +15,17 @@ mongoStoreFactory.$inject = [
 function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
     var exports = {};
 
+    exports.publishRecordByGraphId = function (graphId, event) {
+        return waterline.graphobjects.needOne({ instanceId: graphId })
+            .then(function (graph) {
+                exports.publishGraphRecord(graph, event, graphId);
+            });
+    };
+
+    exports.publishGraphRecord = function(graph, event, id) {
+        return waterline.graphobjects.publishRecord(event || 'updated', graph, id);
+    };
+
     // NOTE: This is meant to be idempotent, and just drop the update silently
     // if the graph has already been marked as done elsewhere and the query returns
     // empty.
@@ -45,7 +56,10 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             new: true
         };
 
-        return waterline.graphobjects.findAndModifyMongo(query, {}, update, options);
+        return waterline.graphobjects.findAndModifyMongo(query, {}, update, options)
+            .tap(function (graph) {
+                exports.publishGraphRecord(graph);
+            });
     };
 
     /**
@@ -100,7 +114,10 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             multi: true
         };
 
-        return waterline.taskdependencies.updateMongo(query, update, options);
+        return waterline.taskdependencies.updateMongo(query, update, options)
+            .tap(function () {
+                exports.publishRecordByGraphId(task.graphId);
+            });
     };
 
     /**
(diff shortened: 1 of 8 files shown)


rolandpoulter:
@RackHD/corecommitters

heckj:
:+1:

anhou:
test this please

yyscamper:
:+1:

lib/workflow/stores/mongo.js
```diff
@@ -283,8 +284,15 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
                 instanceId: 1
             }
         };
+        var promise = waterline.graphobjects.findAndModifyMongo(query, {}, graph, options);
 
-        return waterline.graphobjects.findAndModifyMongo(query, {}, graph, options);
+        if (isNew) {
+            promise.then(function () {
+                waterline.graphobjects.publishRecord('created', graph);
```
yyscamper: Why need the extra publish? As I see the `created` event has been published in model's `afterCreate`.

lib/workflow/stores/mongo.js
```diff
@@ -261,10 +261,11 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
      * Persists a graph to the graphobjects collection
      * @param {Object} graph - the graph object to persist
      * @param {String} graph.instanceId - the unique ID for the graph instance
+     * @param {String} isNew - Boolean, when true a create message is broadcast.
```
yyscamper: `isNew` is Boolean not String.

spec/lib/workflow/task-graph-spec.js
```diff
@@ -498,12 +501,12 @@ describe('Task Graph', function () {
                 definition: definitions.graphDefinition
             })
             .then(function(graph) {
-                return [graph, graph.persist()];
+                return [graph, graph.persist('isNew')];
```
yyscamper: The argument for `persist` is Boolean not String.

yyscamper:
Jenkins: test this please

lib/workflow/stores/mongo.js
benbp: @rolandpoulter I think what we needed here was an 'on.events' message (see https://github.com/RackHD/on-core/blob/master/lib/protocol/events.js#L135) rather than another event on the waterline AMQP exchange?

rolandpoulter: @benbp I've already integrated the waterline events into the UI. We could make this kind of change to move the graph object events to another queue, but for now I'd prefer to get it working again in the UI.

rolandpoulter:
test this please

rolandpoulter:
@benbp need this PR before I can release the operation center fixes to the UI

lib/workflow/stores/mongo.js
```diff
@@ -15,6 +15,17 @@ mongoStoreFactory.$inject = [
 function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
     var exports = {};
 
+    exports.publishRecordByGraphId = function (graphId, event) {
+        return waterline.graphobjects.needOne({ instanceId: graphId })
```
benbp: Isn't this an unnecessary database query @rolandpoulter ?

benbp: Oh, nvm, we need the object.

benbp: Actually, couldn't you just pass the graph object you receive from the `findAndModifyMongo` queries below into this method? Rather than doing this call.

benbp:
@rolandpoulter 👍 with a fix for what I believe are redundant db calls.

rolandpoulter:
@benbp Removed redundant db calls

PR #141, opened 2016-05-17
Fix core/index to include its own default di

rolandpoulter:
This change allows RackHD modules to use the DI from on-core instead of requiring them to have their own DI in package.json:dependencies.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,12 +2,16 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


yyscamper:
:+1:

jlongever:
test this please

benbp:
👍

PR #143, opened 2016-05-17
Refactor DB renderable objects

brianparry:
This patch changes the behavior of DB renderable objects (templates/profiles/views) such that template content is stored on the file system only rather than loaded into the database.

API PUT and DELETE methods will update both the file-system and the template metadata stored in the database.  User changes to templates will be preserved if on-http is restarted.

This change is dependent on https://github.com/RackHD/on-http/pull/293. Tests will not pass until both changes are merged.


brianparry:
Verified functional tests passed when https://github.com/RackHD/on-http/pull/293 is combined with this PR.

lib/common/db-renderable-content.js
```diff
@@ -70,26 +82,84 @@ function dbRenderableContentServiceFactory(
...
+        var filepath = path.join(self.directory, filename);
+        scope = scope || Constants.Scope.Global;
+
+        var promise = new Promise(function(_resolve, _reject) {
+            resolve = _resolve;
+            reject = _reject;
+        });
+
+        stream.on('data', function(chunk) {
+            contents += chunk.toString('utf-8');
```
zyoung51: Could the hash be computed here and the hash passed into loadFile() to avoid reading the file contents twice?

lib/services/lookup.js
```diff
@@ -198,6 +198,10 @@ function lookupServiceFactory(
         };
     };
 
+    LookupService.prototype.reqIpAddressToMacAddress = function (req) {
```
zyoung51: Is this a part of the dbRenderable changes?

brianparry: that's needed for the 2.0 templates route, I figured I'd get the dependency satisfied now, so that I can just submit a single PR for templates.

zyoung51:
+1, Jenkins: test this please

anhou:
+1

PR #145, opened 2016-05-18
Move ARP handler into lookups service

jlongever:
Moves the ARP handler into the lookups services deprecating ARP poller in on-taskgraph.  

Solves a couple known issues with existing ARP poller design:
1. Potential race condition when issuing a POST to /lookups on an IP change,  current ARP poller may see the ARP entry update and update the entry resulting in 500 response on the POST.
2. The existing poller runs as a job in on-taskgraph creating the prerequisite to run on-taskgraph on the same host that is taking the IP request from a client to ensure the hosts ARP table is current.  

@RackHD/corecommitters @uppalk1 @zyoung51 @tannoa2 @keedya

### DIFF
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -7,19 +7,19 @@ module.exports = lookupServiceFactory;
 lookupServiceFactory.$provide = 'Services.Lookup';
 lookupServiceFactory.$inject = [
     'Promise',
-    'Services.Configuration',
     'Services.Waterline',
+    'Services.Configuration',
     'Assert',
     'Errors',
     '_',
-    'lru-cache'
+    'lru-cache',
+    'ARPCache'
 ];
 
 /**
  * Provides a LookupService singleton.
  * @private
  * @param  {Promise} Promise
- * @param  {configuration} configuration
  * @param  {Waterline} waterline
  * @param  {Assert} assert
  * @param  {Errors} errors
@@ -30,12 +30,13 @@ lookupServiceFactory.$inject = [
  */
 function lookupServiceFactory(
     Promise,
-    configuration,
     waterline,
+    configuration,
     assert,
     Errors,
     _,
-    LRU
+    LRU,
+    arpCache
 ) {
     // TODO: stolen from http-service could stand to be a re-usable helper.
     /**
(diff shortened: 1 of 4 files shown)


jlongever:
test this please

yyscamper:
@jlongever : so I would expect, with your code, every lookup CURD operation need to firstly analyze the current ARP data?

jlongever:
@yyscamper only on a LRU cache miss.

jlongever:
Unit-test failures for on-http requires https://github.com/RackHD/on-http/pull/300

yyscamper:
@jlongever , thx. :+1:

benbp:
test this please

lib/services/lookup.js
```diff
@@ -317,27 +331,28 @@ function lookupServiceFactory(
...
+                    return self.assignNodeIdCache(ip, record.node);
+                }
+                else {
+                    self.clearNodeIdCache(ip);
+                    throw new Errors.NotFoundError(
+                        'Lookup Record Not Found (ipAddressToNodeId)',
+                        record
+                    );
+                }
+            }.bind(self)).catch(self.clearNodeIdCacheAndThrow.bind(self, ip));
```
benbp: Binding to `self` is redundant if you are using `self` references within the callback. If you wanted to use `this` references, then it would make more sense (as some of the other code appears to be doing). It's inconsequential, but slightly confusing.

benbp:
👍

benbp:
Why not check the cache for all conversion calls too (such as macAddressToNode, etc.)?

lib/services/lookup.js
jlongever: This is the original code, diff is indent. But you're right. Thanks.

jlongever:
@benbp the LRU cache?  The original code never included it, maybe it was overlooked then?

lib/services/lookup.js
jlongever: Bah, nvm. I changed it to self and left the bind as-is.

jlongever:
test this please

jlongever:
@benbp are you OK with this being merged?

benbp:
@jlongever I mean like there seems to be no reason to call `validateArpCache` when doing things like `macAddressToNode`, it's just unnecessary I/O right?

benbp:
Sorry for the delay :)

jlongever:
@benbp updated.

jlongever:
test this please

benbp:
👍

PR #146, opened 2016-05-24
Generate event when tags are added/removed

zyoung51:
and complete the TODO of changing to findAndModifyMongo command

### DIFF
--- a/lib/models/node.js
+++ b/lib/models/node.js
@@ -10,10 +10,11 @@ NodeModelFactory.$inject = [
     'Services.Waterline',
     '_',
     'Promise',
-    'Constants'
+    'Constants',
+    'Protocol.Waterline'
 ];
 
-function NodeModelFactory (Model, waterline, _, Promise, Constants) {
+function NodeModelFactory (Model, waterline, _, Promise, Constants, waterlineProtocol) {
     return Model.extend({
         connection: 'mongo',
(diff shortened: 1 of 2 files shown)


lib/models/node.js
```diff
@@ -82,28 +83,34 @@ function NodeModelFactory (Model, waterline, _, Promise, Constants) {
...
                             $pull: { tags: tag },
                             $set: { updatedAt: new Date() }
-                        });
-                });
+                        };
+            var options = { new: true };
+            return self.findAndModifyMongo(query, {}, update, options)
+            .then(function(record) {
+                record.id = record._id;
+                waterlineProtocol.publishRecord(self, 'updated', record)
```
benbp: Wonder if we should just move the publishRecord code into the `findAndModifyMongo` function?

benbp:
👍  after semicolons.

jlongever:
+1

lib/models/node.js
yyscamper: The event should be "remove" rather than "updated"

yyscamper: The 'event' should be 'created' rather than 'updated'

yyscamper: ahh, for node, the event is 'updated', but do you think whether we need to define the event to explicit 'tag added' rather than 'node updated'?

PR #149, opened 2016-05-31
add constants for drive secure erase

iceiilin:
The tools and corresponding parameters (named "operation" here) varies for different type of drive's protocols. This PR adds support list about the relationship of "tool - protocol - operation".

Related PR:
https://github.com/RackHD/on-tasks/pull/225


benbp:
Are these values used anywhere other than the job in on-tasks? If not, I'm not sure it makes sense to add them to constants.

iceiilin:
Move drive const to job. close this PR.

PR #164, opened 2016-06-17
Replace node.obmSettings with obms collection.

brianparry:
* Create new obms model and tests.
* Remove obmSettings property from nodes resource.
* Add 1-N relation from nodes to OBMs

Supports:
https://github.com/RackHD/on-http/pull/312
https://github.com/RackHD/on-tasks/pull/243
https://github.com/RackHD/on-http/pull/334

@RackHD/corecommitters

### DIFF
--- a/lib/models/node.js
+++ b/lib/models/node.js
@@ -29,10 +29,9 @@ function NodeModelFactory (Model, waterline, _, Promise, Constants, configuratio
             type: 'string',
             required: true
         },
-        obmSettings: {
-            type: 'json',
-            required: false,
-            json: true
+            obms: {
+                collection: 'obms',
+                via: 'node'
         },
         type: {
             type: 'string',
--- a/lib/models/obms.js
+++ b/lib/models/obms.js
@@ -1,0 +1,183 @@
+// Copyrighr 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = ObmsModelFactory;
+
+ObmsModelFactory.$provide = 'Models.Obms';
+ObmsModelFactory.$inject = [
+    'Model',
+    'Services.Waterline',
+    '_',
+    'Services.Encryption',
+    'Constants',
+    'Promise',
+    'util',
+    'Errors',
+    'Assert'
+];
+
+function ObmsModelFactory (
+    Model,
+    waterline,
+    _,
+    encryption,
+    Constants,
+    Promise,
+    util,
+    Errors,
+    assert
+) {
+    var secretsPattern = _.reduce(Constants.Logging.Redactions, function(pattern, regex) {
+        return pattern ? util.format('%s|(?:%s)', pattern, regex.source) :
+                         util.format('(?:%s)', regex.source);
+    }, '');
+    var secretsRegex = new RegExp(secretsPattern, 'i');
+
+    function _encryptSecrets(obj) {
+        assert.object(obj);
+        assert.object(obj.config);
+
+        obj.config = _(obj.config).mapValues(function(value, key) {
+            if (key.match(secretsRegex)) {
+                return encryption.encrypt(value);
+            }
+            return value;
+        }).value();
+
+        return obj;
+    }
+
+    function _deleteSecrets(obj) {
+        assert.object(obj);
+        assert.object(obj.config);
+
+        obj.config = _(obj.config).omit(function(value, key) {
+            return key.match(secretsRegex);
+        }).value();
+
+        return obj;
+    }
+
+    function _revealSecrets(obj) {
+        assert.object(obj);
+        assert.object(obj.config);
+
+        _.forEach(obj.config, function(value, key) {
+            if (key.match(secretsRegex)) {
+                obj.config[key] = encryption.decrypt(value);
+            }
+        });
+
+        return obj;
+    }
+
+    return Model.extend({
+        connection: 'mongo',
+        identity: 'obms',
+        attributes: {
+            node: {
+                model: 'nodes',
+                required: true
+            },
+            service: {
+                type: 'string',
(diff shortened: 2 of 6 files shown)


lib/models/obms.js
```diff
@@ -0,0 +1,183 @@
...
+            var self = this;
+
+            return waterline.nodes.findByIdentifier(nodeId)
+            .then(function(nodeRecord) {
+                node = nodeRecord;
+                query = {node: node.id, service: obm.service};
+                return self.find(query);
+            })
+            .then(function(obmSettings) {
+                assert.ok(obmSettings.length === 0 || obmSettings.length === 1);
```
benbp: An assert description as the second argument here could be really useful for debug.

lib/models/obms.js
```diff
@@ -0,0 +1,183 @@
...
+        //    between the atomic create or modify and find operations.  This
+        //    will cause the function fulfillment value to be inconsistent
+        //    with the provided obm arguments but consistent with the document
+        //    content at the time of the find query.
+        upsertByNode: function(nodeId, obm) {
+            var node;
+            var query;
+            var self = this;
+
+            return waterline.nodes.findByIdentifier(nodeId)
```
benbp: Is this `findByIdentifier` call necessary? Seems like all that's being used is `node.id` which should be equal to `nodeId` anyways.

srinia6:
+1

brianparry:
@benbp I pushed a new patch to address your comments.

lib/models/node.js
```diff
@@ -29,10 +29,9 @@ function NodeModelFactory (Model, waterline, _, Promise, Constants, configuratio
             type: 'string',
             required: true
         },
-        obmSettings: {
-            type: 'json',
-            required: false,
-            json: true
+            obms: {
```
anhou: Indent

lib/models/obms.js
```diff
@@ -0,0 +1,183 @@
...
+
+                _.merge(query, {service: service, node: node.id} );
+                return self.findOne(query);
+            })
+            .then(function(obm) {
+                if (obm && reveal) {
+                    return _revealSecrets(obm);
+                }
+                return obm;
+            })
```
anhou: need to be fixed

anhou:
:+1:

lib/models/obms.js
```diff
@@ -0,0 +1,183 @@
...
+    waterline,
+    _,
+    encryption,
+    Constants,
+    Promise,
+    util,
+    Errors,
+    assert
+) {
+    var secretsPattern = _.reduce(Constants.Logging.Redactions, function(pattern, regex) {
```
yyscamper: Actually the `Redactions` can be either regex or string, the string type redaction means "full word matching", check this code: https://github.com/RackHD/on-core/blob/master/lib/serializables/log-event.js#L149-L161

I think the best method is to move the Redaction to a common service function, so that it can be leveraged by other modules.  I can accept to do it with another PR.

lib/models/obms.js
```diff
@@ -0,0 +1,183 @@
...
+        //    between the atomic create or modify and find operations.  This
+        //    will cause the function fulfillment value to be inconsistent
+        //    with the provided obm arguments but consistent with the document
+        //    content at the time of the find query.
+        upsertByNode: function(nodeId, obm) {
+            var query;
+            var self = this;
+
+            return Promise.try(function() {
+                query = {node: nodeId, service: obm.service};
```
yyscamper: need some assertion logic for `obm` and `obm.service`

lib/models/obms.js
brianparry: @yyscamper Redactions would actually be easier to aggregate if they were strings, not patterns, so I'm all for changing this.  I don't really think there's a use case for taking them individually, so I'd vote for structuring them so that they are easy to merge.

As for a common redaction service, I think this is unnecessary if encryption is managed in the models.  My preference is to make each model responsible for encrypting any secrets it might contain since models are responsible for their own content.  Encryption is a form of redaction since plain text secrets are replaced with a cipher.  I also think that `toJSON()` should remove any secret properties from the model before it is used in logs, views, etc since redacted data has no value to the user.

lib/models/obms.js
brianparry: @yyscamper This schema validator should already have rejected an invalid service before getting here.

yyscamper: OK, didn't realized the model schema.

lib/models/obms.js
brianparry: @yyscamper Actually, I think it would be even better if we a custom validation called `secret` at the model level.  That would allow us to make secrets management totally generic within the models.  If a model has a password property, rather than a regex match we'd defiine it like so in the model:

```
password: {
      type: 'string',
      secret: true,
}
```

yyscamper: The `secret` is a good idea, but it can only cover the case that we have solid key in model.  Take the Task options for example, the task model only defines a key `options`, but the properties in `options` may still contain some sensitive data, like the rootPassword in install-os task. You may think that we can extend this key in task schema as well. But seems the task model is hard to access the external task schema.

brianparry: @yyscamper That's true, but it just reveals a problem with the way our models our structured.  In many places (including obms) we define model properties with `type: 'json'` and lose all power to place constraints and/or validations around that data at the model level.  In the case of obms, we have an external schema validator to help with this, but I think in most cases it indicates that we need to rethink the model design.  This problem is further complicated by a waterline limitation: https://github.com/balderdashy/sails-mongo/issues/44.

yyscamper:
:+1:

PR #165, opened 2016-06-20
Create JsonSchemaValidator using Ajv lib

WangWinson:
This is a base class for json schema validation.
As  @yyscamper 's design.
- Ajv library should be leveraged to validate JSON schema
- In this initial version, expose `validate`, `addSchema`, `getSchema` function
- Detail implement of skip `context` validate will be in the subclass in on-tasks

@RackHD/corecommitters

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,7 +51,7 @@ module.exports = function (di, directory) {
             helper.requireWrapper('pluralize', 'pluralize'),
             helper.requireWrapper('always-tail', 'Tail'),
             helper.requireWrapper('flat', 'flat'),
-            helper.requireWrapper('ajv', 'ajv'),
+            helper.requireWrapper('ajv', 'Ajv'),
 
             // Glob Requirables
             helper.requireGlob(__dirname + '/lib/common/*.js'),
--- a/index.js
+++ b/index.js
@@ -51,6 +51,7 @@ module.exports = function (di, directory) {
             helper.requireWrapper('pluralize', 'pluralize'),
             helper.requireWrapper('always-tail', 'Tail'),
             helper.requireWrapper('flat', 'flat'),
+            helper.requireWrapper('ajv', 'ajv'),
 
             // Glob Requirables
             helper.requireGlob(__dirname + '/lib/common/*.js'),
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -6,14 +6,14 @@ module.exports = jsonSchemaValidatorFactory;
 
 jsonSchemaValidatorFactory.$provide = 'JsonSchemaValidator';
 jsonSchemaValidatorFactory.$inject = [
-    'ajv'
+    'Ajv'
 ];
 
 function jsonSchemaValidatorFactory(
     Ajv
 ) {
-    function JsonSchemaValidator () {
-        this._ajv = new Ajv({ allErrors: true,  verbose: true});
+    function JsonSchemaValidator(options) {
+        this._ajv = new Ajv(options || {});
         this.addSchema = this._ajv.addSchema;
         this.addMetaSchema = this._ajv.addMetaSchema;
     }
@@ -28,7 +28,7 @@ function jsonSchemaValidatorFactory(
         if (this._ajv.validate(schema, data)) {
             return true;
         } else {
-            var err = new Error('JSON schema validtion failed - ' + this._ajv.errorsText());
+            var err = new Error('JSON schema validation failed - ' + this._ajv.errorsText());
             err.errorList = this._ajv.errors;
             throw err;
         }
@@ -36,7 +36,7 @@ function jsonSchemaValidatorFactory(
 
     /**
      * Get compiled schema from the instance by `key` or `ref`.
-     * @param  {String} keyRef `key` that was passed to `addSchema` or 
+     * @param  {String} key `key` that was passed to `addSchema` or
      *                  full schema reference (`schema.id` or resolved id).
      * @return {Object} schema
      */
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -1,0 +1,51 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = jsonSchemaValidatorFactory;
+
+jsonSchemaValidatorFactory.$provide = 'JsonSchemaValidator';
+jsonSchemaValidatorFactory.$inject = [
+    'ajv'
+];
+
+function jsonSchemaValidatorFactory(
+    Ajv
+) {
+    function JsonSchemaValidator () {
+        this._ajv = new Ajv({ allErrors: true,  verbose: true});
+        this.addSchema = this._ajv.addSchema;
+        this.addMetaSchema = this._ajv.addMetaSchema;
+    }
+
+    /**
+     * validate JSON data with given JSON schema
+     * @param  {Object|String} schema  JSON schema Object or schema ref id
+     * @param  {Object} data  JSON data to be validated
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -51,6 +51,7 @@ module.exports = function (di, directory) {
             helper.requireWrapper('pluralize', 'pluralize'),
             helper.requireWrapper('always-tail', 'Tail'),
             helper.requireWrapper('flat', 'flat'),
+            helper.requireWrapper('ajv', 'ajv'),
```
yyscamper: Change the injectable Name to `Ajv` will give clear message for user that it is a Class not instance.

lib/common/json-schema-validator.js
```diff
@@ -0,0 +1,51 @@
...
+        } else {
+            var err = new Error('JSON schema validtion failed - ' + this._ajv.errorsText());
+            err.errorList = this._ajv.errors;
+            throw err;
+        }
+    };
+
+    /**
+     * Get compiled schema from the instance by `key` or `ref`.
+     * @param  {String} keyRef `key` that was passed to `addSchema` or
```
yyscamper: By your function design, the param name is `key` not `keyRef`

lib/common/json-schema-validator.js
```diff
@@ -0,0 +1,51 @@
...
+jsonSchemaValidatorFactory.$provide = 'JsonSchemaValidator';
+jsonSchemaValidatorFactory.$inject = [
+    'ajv'
+];
+
+function jsonSchemaValidatorFactory(
+    Ajv
+) {
+    function JsonSchemaValidator () {
+        this._ajv = new Ajv({ allErrors: true,  verbose: true});
```
yyscamper: Other schema validator may use special configuration, for example redfish-schema-validator doesn't need the "allErrors: true" as it doesn't need the special "context" validation.
So it's better to let the Validator supports additional options:
```
function JsonSchemaValidator(options) {
   this._ajv = new Ajv(options || {allErrors: true, verbose:  true});
}
```

Or we can completely remove the default option:
```
function JsonSchemaValidator(options) {
   this._ajv = new Ajv(options || {});
}
```

lib/common/json-schema-validator.js
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Get compiled schema from the instance by `key` or `ref`.
+     * @param  {String} keyRef `key` that was passed to `addSchema` or 
+     *                  full schema reference (`schema.id` or resolved id).
+     * @return {Object} schema
+     */
+    JsonSchemaValidator.prototype.getSchema = function (key) {
+        var schemaCompiled = this._ajv.getSchema(key);
+        if (schemaCompiled) {
```
yyscamper: Please confirm whether the returned schema has resolved all schema reference?
According our discussion, we should support both kinds of following schemas:
(1) The schema that has resolved all references.
(2) The raw schema definition, no reference is resolved

spec/lib/common/json-schema-validator-spec.js
```diff
@@ -0,0 +1,47 @@
...
+
+    it('should return true when validation pass', function () {
+        var result = validator.validate(testSchema1, { repo : '/172.31.128.1/mirrors' });
+        expect(result).to.be.true;
+    });
+
+    it('should throw validation error with incorrect data format', function () {
+        expect(function () {
+            validator.validate(testSchema1, { repo : 'abc' });
+        }).to.throw(/JSON schema validtion failed/);
```
yyscamper: typo? validtion => validation

spec/lib/common/json-schema-validator-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    it('should add correct schema', function () {
+        testSchema1.id = 'test1';
+        expect(validator.addSchema(testSchema1)).to.be.empty;
+    });
+
+    it('should get existing schema', function () {
+        expect(validator.getSchema('test1')).to.deep.equal(testSchema1);
```
yyscamper: If the getSchema can resolve all JSON references, it's better to add a test case to check this feature.

spec/lib/common/json-schema-validator-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    it('should not find schema not exist', function () {
```
yyscamper: This test case and above 2 cases have inter-dependency, actually this is not a good practice. A single test should not depend on running other tests before it, nor should it be affected by the order of execution of other tests.

If you want to control the unit-test execution, you can turn to use 'before/beforeEach/after/afterEach', you should not expect the unit-test will be run just as your lexical order.

lib/common/json-schema-validator.js
WangWinson: Agreed, Since this is base class, option 2 maybe better.  The specific option would be inside the sub class

spec/lib/common/json-schema-validator-spec.js
WangWinson: OK, I'll add the test case

index.js
WangWinson: make sense. updated

lib/common/json-schema-validator.js
WangWinson: you knew it, copy & paste issue. lol

spec/lib/common/json-schema-validator-spec.js
WangWinson: sharp eye!

yyscamper: This throw another question, since you had the typo but the mocha doesn't catch it! Is their any usage problem about chai's `throw`?

WangWinson: It's because I had typo in the code first , then copy & paste the string to unit test ...

yyscamper: ^^

WangWinson:
test this please

lib/common/json-schema-validator.js
WangWinson: As investigated, Ajv don't support return a reference resolved schema directly. Additional effort needed to implement a `getSchemaResolved` function.
This will be added in another PR later.

yyscamper: OK

yyscamper:
:+1:

lib/common/json-schema-validator.js
```diff
@@ -0,0 +1,51 @@
...
+     * Get compiled schema from the instance by `key` or `ref`.
+     * @param  {String} key `key` that was passed to `addSchema` or
+     *                  full schema reference (`schema.id` or resolved id).
+     * @return {Object} schema
+     */
+    JsonSchemaValidator.prototype.getSchema = function (key) {
+        var schemaCompiled = this._ajv.getSchema(key);
+        if (schemaCompiled) {
+            return schemaCompiled.schema;
+        }
```
iceiilin: As a base class, do we need to throw something if no matched schema? so that the upper class doesn't need to do special check.

iceiilin:
:+1:

WangWinson:
test this please

lib/common/json-schema-validator.js
WangWinson: Good question.  
It depends on how would the upper class would like to use it. Most likely there will be some logic to handle this. For example, this will be called in /api/schema/<schema_name> API router in future. It still need to translate the result/error to an HTTP error 404 maybe. 
So I would like to leave this as-is so far until we found that we did need an error throw during real use case.

cgx027:
:+1:

WangWinson:
test this please

yyscamper:
:+1:

PR #166, opened 2016-06-22
Update Task/TaskGraph AMQP messenger plugin to have big timeout.

jimturnquist:
Based on the build machine, the Task/TaskGraph AMQP messenger plugin can take more than the default allotment of 2000ms.  The timeout was upped significantly to 20000ms.

Ran this on our local build machine that was experiencing the failures to verify.

@RackHD/corecommitters @benbp @jlongever @heckj

### DIFF
--- a/HWIMO-TEST
+++ b/HWIMO-TEST
@@ -13,9 +13,9 @@ npm install
 ./node_modules/.bin/jshint -c .jshintrc lib index.js || true
 
 # run the tests without redirected output for debugging
-./node_modules/.bin/_mocha $(find spec -name '*-spec.js') --require spec/helper.js
+./node_modules/.bin/_mocha $(find spec -name '*-spec.js') --timeout 10000 --require spec/helper.js
 
 # Runs the mocha tests and reports the code coverage.
-./node_modules/.bin/istanbul cover -x "**/spec/**" ./node_modules/.bin/_mocha -- $(find spec -name '*-spec.js') -R xunit-file --require spec/helper.js
+./node_modules/.bin/istanbul cover -x "**/spec/**" ./node_modules/.bin/_mocha -- $(find spec -name '*-spec.js') --timeout 10000 -R xunit-file --require spec/helper.js
 ./node_modules/.bin/istanbul report cobertura
 


benbp:
Is it the `afterEach` function that's timing out? Or something like the `before` function?

jimturnquist:
The "before all" hook is timing out.

benbp:
Are we sure this isn't just whack-a-mole, and we'll hit a timeout issue with another unit test later? I ask because the only thing `before` does here is load a single file from disk, and create a few mock objects. At first glance doesn't seem like we're doing anything complex that would take more than 2 seconds. It seems to me it could either be an issue with the test harness or with the test/build environment.

The change is harmless, so there's little reason not to merge it, but I'm not convinced it will address the real problem that you are having.

jimturnquist:
Thanks Ben.  Yes, adding timeouts is always a game of whack-a-mole.  We are hitting this particular issue probably 90% of our test runs when building on-core.  Here is a summary of what is being done during this particular 'before' function.  The time appears to be spent in Setup injector.

JAT: Setup injector: 2119ms
JAT: Protocol.Task: 0ms
JAT: Protocol.Events: 0ms
JAT: Protocol.TaskGraphRunner: 0ms
JAT: Task.Messengers.AMQP: 0ms

yyscamper:
@benbp @jimturnquist :
This is test suite-level timeout, that means all test cases in this test suite should be completed within that amount of time. So it's not the `before` causes the timeout, it's all of those test cases.

Actually, I think the unit-test itself should not touch the real rabbitmq service, that's not what unit-test should not.

jimturnquist:
@yyscamper Please look at the timings from my previous post.  All of those timings are coming from the 'before' function.  

The culprit is:
JAT: Setup injector: 2119ms

Without this change, the following can be seen reported by the test:

Task/TaskGraph AMQP messenger plugin "before all" hook:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

yyscamper:
@jimturnquist : Sorry, I feel confused, if it is due to the `before all` hook, then why increase the suite-level timeout? Increase the timeout for `before` hook is enough I think.

jimturnquist:
Its my understanding that the before hook is included in the suite-wide timing.  Setting on just the before hook won't be adequate.

jimturnquist:
@RackHD/corecommitters 

The Corsair team is waiting on a stable Jenkins build of on-core in order for other packaging dependencies to happen.  Any attention given to this PR would be appreciated.

yyscamper:
@jimturnquist : could you please decrease the timeout value to a lower one? I think the 20000 is so large.

Though I think purely changing the timeout will hide some real problem about unit-test, but I know @PengTian0 also complain the on-core timeout issue in the new Jenkins CI server. So I'm OK to merge this PR first as a workaround, because I guess it will not be an easy task to completely find the cause and fix the unit-test defects.

So :+1: after lower the timeout amount.

jimturnquist:
Interesting.  The following within setupInjector is taking almost all of the time:

    setupInjector: function (overrides) {
        // Start with the core dependencies.
+        console.time('JAT: require');
        var core = require('../index')(di, '..');
+        console.timeEnd('JAT: require');


JAT: require: 2143ms

jimturnquist:
I will decrease the timeout down to 10000ms at the request of @yyscamper.  I don't want to go much less  as I have seen timings over 5s before.

jimturnquist:
Timer has been lowered to 10s.

jimturnquist:
I have respun this fix so that the timeout is only specified in HWIMO-TEST.

benbp:
👍 Thanks @jimturnquist

benbp:
@yyscamper I did some preliminary investigation with strace, but nothing glaringly obvious stood out to me. @jimturnquist sent me [this article](https://kev.inburke.com/kevin/node-require-is-dog-slow/), and I've been finding a lot of missed locations looking for kerberos related dependencies, but I haven't identified why, for example, tests pass on my local, very slow VM vs. a powerful Jenkins one (as an anecdote though, the file I/O tends to be faster on my local VM, and another difference is I installed with npm v2 whereas the problem machine is using npm v3). I think we can use the `--timeout` flag in HWIMO-TEST as a workaround that will fix the issue for Jim without requiring _everybody_ to use a bigger timeout, and give us some more time to figure out more definitively what the root cause is (I'm still largely suspecting something about the OS).

Happy to share the strace logs from failing and working builds if anyone is interested in taking a look.

PR #170, opened 2016-06-30
add getSchemaResolved function

WangWinson:
While `getSchema` function return the original schema added to the validator instance,
`getSchemaResolved` is to return a schema with reference resolved.
example:
```JSON
{
        "id": "/refschema/r1",
        "definitions": {
            "url": {
                "type": "string",
                "format": "uri"
            }
        },
        "properties“: {
            ”repo": {
                "$ref": "#/definitions/url"
            }
        }
};
``` 
after resolve
```JSON
{
        "id": "/refschema/r1",
        "definitions": {
            "url": {
                "type": "string",
                "format": "uri"
            }
        },
        "properties": {
            "repo": {
                "type": "string",
                "format": "uri"
            }
        }
};
```
 
@RackHD/corecommitters @cgx027 @iceiilin

### DIFF
--- a/spec/lib/common/json-schema-validator-spec.js
+++ b/spec/lib/common/json-schema-validator-spec.js
@@ -41,6 +41,172 @@ describe('JsonSchemaValidator', function () {
         }
     };
 
+    var testRefSchema1Resolved = {
+        id: '/refschema/r1',
+        definitions: {
+            url: {
+                type: 'string',
+                format: 'uri'
+            }
+        },
+        properties: {
+            repo: {
+                type: 'string',
+                format: 'uri'
+            },
+            index: {
+                type: 'number',
+            }
+        }
+    };
+
+    var testRefSchema2Resolved = {
+        id: '/refschema/r2',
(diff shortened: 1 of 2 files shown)


iceiilin:
+1

spec/lib/common/json-schema-validator-spec.js
```diff
@@ -92,5 +258,18 @@ describe('JsonSchemaValidator', function () {
             validator.getSchema('/refschema/r2');
         }).to.throw(/can't resolve reference /);
     });
+
+    it('should get schema with reference resolved', function () {
```
yyscamper: Suggest adding one more test case:
 - no $ref in the schema, make sure the getSchemaResolved behaves well. Because sometimes the code may work for very complex cases but fail in the very simple case, this case is to make sure it works in both simple & complex cases.

yyscamper:
Still :+1: for me.

WangWinson:
Thanks @yyscamper , added test case as your suggestion

PR #171, opened 2016-06-30
json-schema-validator: add validatePatternSkipped & reset function

yyscamper:
add functions to json-schema-validator:

1 - `validatePatternSkipped`: validate the data against schema, but skip some validation if the error matches the skip-patterns. This is to support task schema upfront validation, which need to skip all context template.
2. `reset`: reset the validator to initial state.

@RackHD/corecommitters @WangWinson @iceiilin @cgx027 @pengz1

### DIFF
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -7,17 +7,20 @@ module.exports = jsonSchemaValidatorFactory;
 jsonSchemaValidatorFactory.$provide = 'JsonSchemaValidator';
 jsonSchemaValidatorFactory.$inject = [
     'Ajv',
+    'Assert',
     '_'
 ];
 
 function jsonSchemaValidatorFactory(
     Ajv,
+    assert,
     _
 ) {
     var url = require('url');
 
     function JsonSchemaValidator(options) {
-        this._ajv = new Ajv(options || {});
+        this.options = options || {};
+        this._ajv = new Ajv(this.options);
         this.addSchema = this._ajv.addSchema;
         this.addMetaSchema = this._ajv.addMetaSchema;
     }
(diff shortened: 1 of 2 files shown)


lib/common/json-schema-validator.js
```diff
@@ -35,6 +40,52 @@ function jsonSchemaValidatorFactory(
...
+            return true;
+        }
+
+        var errors = this._ajv.errors;
+
+        if (skipPatterns) {
+            if (!_.isArray(skipPatterns)) {
+                skipPatterns = [skipPatterns];
+            }
+            assert.arrayOfRegexp(skipPatterns, 'skip pattern should be regexp');
```
WangWinson: Is it reasonable to throw `errors` directly when illegal `skipPatterns` passed ?

lib/common/json-schema-validator.js
```diff
@@ -35,6 +40,52 @@ function jsonSchemaValidatorFactory(
...
+
+            errors = _.filter(errors, function(error) {
+                return !_.some(skipPatterns, function(pattern) {
```
WangWinson: so, every regex in the `skipPatterns` array is the `&&` relationship?

lib/common/json-schema-validator.js
```diff
@@ -47,5 +98,9 @@ function jsonSchemaValidatorFactory(
         }
     };
 
+    JsonSchemaValidator.prototype.reset = function () {
```
WangWinson: Well, may need js doc style comments here

spec/lib/common/json-schema-validator-spec.js
```diff
@@ -53,44 +53,173 @@ describe('JsonSchemaValidator', function () {
 
     helper.after();
 
-    it('should return true when validation pass', function () {
-        var result = validator.validate(testSchema1, { repo : '/172.31.128.1/mirrors' });
-        expect(result).to.be.true;
+    describe('validate', function() {
```
WangWinson: I have separated the `describe` section for each method. So there will be conflict to resolve here. (if my PR goes in first)

spec/lib/common/json-schema-validator-spec.js
```diff
@@ -53,44 +53,173 @@ describe('JsonSchemaValidator', function () {
...
+            expect(function () {
+                validator.getSchema('/refschema/r2');
+            }).to.throw(/can't resolve reference /);
+        });
     });
 
-    it('should not find not existing schema', function () {
-        expect(validator.getSchema('test2')).to.be.undefined;
+    describe('validatePatternsSkipped', function() {
+        var skipPatterns = [ /{{\s*foo.\w+\s*}}/, /<%\s*bar.\d+\s*%>/ ];
```
WangWinson: If you do want to match brace `{` sth `}` here. Should use backslash to escape char. `\{\{`

lib/common/json-schema-validator.js
yyscamper: not AND but OR relationship, if any of pattern matches then the error is ignored.

yyscamper: This expression means: if the error doesn't match any of the skipPatterns, the filter it out to the result.

lib/common/json-schema-validator.js
yyscamper: if skipPatterns is in invalid format, but we skip it and consider it as no skipPatterns, then user will not be aware of such error. I think throw the assertion error is better here.

spec/lib/common/json-schema-validator-spec.js
yyscamper: I hate resolve conflict. But that's done^^

spec/lib/common/json-schema-validator-spec.js
yyscamper: Good catch. But `<` doesn't need be escaped, and even the jshint will give warning if it is escaped.

yyscamper:
@WangWinson , I have addressed your comment.

anhou:
:+1:

WangWinson:
Thanks  
:+1:

benbp:
👍

PR #172, opened 2016-07-01
Create alert for poller status changes

pengz1:
This PR is to add alert capability in workitem .setFailed and .setSucceededwhen methods when pollers status changed and it is decided node alert should be issued.

A new argument is inserted in .setFailed and .setSucceded method and old tasks might be impacted and need update.

Code is Verified on D51, vPDU.

Related PR:
https://github.com/RackHD/on-tasks/pull/262

### DIFF
--- a/lib/models/work-item.js
+++ b/lib/models/work-item.js
@@ -82,7 +82,7 @@ function WorkItemModelFactory (Model, _, Promise, assert, Constants, configurati
                 type: 'boolean',
                 defaultsTo: false
             },
-            status:{
+            state:{
                 type: 'string',
                 defaultsTo: null
             },
@@ -151,16 +151,16 @@ function WorkItemModelFactory (Model, _, Promise, assert, Constants, configurati
 
         setFailed: function setFailed(leaseToken, alertMessage, workItems) {
             var self = this;
-            var newStatus = "inaccessible";
+            var newState = "inaccessible";
             if (!Array.isArray(workItems)) {
                 workItems = Array.prototype.slice.call(arguments, 2);
             }
             var now = new Date();
             return Promise.all(_.map(workItems, function (workItem) {
                 if (!alertMessage) {
-                    newStatus = workItem.status;
-                } else {
-                    accessibleAlert(alertMessage, workItem, newStatus);
+                    newState = workItem.state;
+                } else if (! _.isEmpty(alertMessage)){
+                    accessibleAlert(alertMessage, workItem, newState);
                 }
                 return self.update({
                     id: workItem.id,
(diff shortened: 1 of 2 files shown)


pengz1:
@RackHD/corecommitters @cgx027

lib/models/work-item.js
```diff
@@ -246,4 +265,37 @@ function WorkItemModelFactory (Model, _, Promise, assert, Constants, configurati
...
+            var alertInfo = _.assign(
+                {
+                    nodeId: workitem.node,
+                    status: "accessible",
+                    pollerType: workitem.type,
+                    configure: workitem.config
+                },
+                alertMessage
+            );
+            return events.publishNodeAlert(workitem.node, alertInfo);
```
benbp: Might be good to have a `.catch` handler here that will log any failures.

lib/models/work-item.js
```diff
@@ -246,4 +265,37 @@ function WorkItemModelFactory (Model, _, Promise, assert, Constants, configurati
...
+        }
+    }
+
+    function inaccessibleAlert(alertMessage, workitem) {
```
anhou: inaccessibleAlert and accessibleAlert could combine to one function

cgx027:
Looks good to me :+1:

brianparry:
test this please

brianparry:
test this please

brianparry:
test this please

pengz1:
test this please.

pengz1:
This PR has inter-dependency with below PR:
https://github.com/RackHD/on-tasks/pull/262
We need merge this PR first and then to merge on-tasks PR262.

If any issue we can revert code back.

lib/models/work-item.js
```diff
@@ -144,13 +149,19 @@ function WorkItemModelFactory (Model, _, Promise, assert, Constants, configurati
             });
         },
 
-        setFailed: function setFailed(leaseToken, workItems) {
+        setFailed: function setFailed(leaseToken, alertMessage, workItems) {
```
WangWinson: Any reason why the `alertMessage` param added in middle of the original params ?  This way you need to make sure to update everywhere the `setFailed` function called.
 
Would it be better to add this as the last param ? so that those code where call the function will still work if you miss to update them.

pengz1: Thanks for the comments. I put alertMessage to make it convenient to slice possible workItems array.
I already update all setFailed so suggest to keep as it is.

yyscamper:
jenkins: test this please

yyscamper:
:+1:

yyscamper:
This PR changes some API of on-core, so if the Jenkins build use the old on-tasks, the smoke testing will fail. I need to merge this PR first regardless of the Jenkins failure, then trigger the testing of that dependant on-tasks.

PR #174, opened 2016-07-04
Fix nested ref not resolved issue for incorrect baseId

WangWinson:
The issue was found when working on task annotation.
Since the baseId (`rackhd/schemas/v1/tasks/xxxxx`) scope changed, some $ref not been replaced correctly.

@RackHD/corecommitters @cgx027 @iceiilin

### DIFF
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -190,16 +190,16 @@ function jsonSchemaValidatorFactory(
                 }
             }
 
-            for (var k in obj) {
-                if (!(obj[k] instanceof Object)) {
-                    continue;
+            _.forOwn(obj, function(val, k) {
+                if (!(val instanceof Object)) {
+                    return true; //continue
                 }
 
-                var resolvedObj = replaceRefObj(obj[k], resolvedValues, baseId, replaced);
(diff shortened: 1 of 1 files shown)


yyscamper:
:+1:

lib/common/json-schema-validator.js
```diff
@@ -176,19 +181,20 @@ function jsonSchemaValidatorFactory(
...
             if (obj.$ref) {
                 var refId = url.resolve(baseId, obj.$ref);
                 if (refId in resolvedValues) {
-                    replaced.push(refId);
+                    replaced[refId] = resolvedValues[refId];
                     return resolvedValues[refId];
                 }
             }
 
             for (var k in obj) {
```
brianparry: This will iterate over properties in the prototype chain as well.  Is that the intended behavior?

brianparry:
test this please

yyscamper:
Jenkins: test this please

lib/common/json-schema-validator.js
WangWinson: Good question. I know this maybe a concern. But the schema object is defined by ourselves, so it's surely that there will be no additional properties in the prototype chain.

iceiilin:
+1

PR #176, opened 2016-07-08
Add check graph succeeded method

benbp:
Adding basically a copy of the `checkGraphFinished` store method, but instead of querying for pending tasks, we query for pending AND failed tasks (i.e. != succeeded). This allows a higher level of the stack to prevent succeeding a graph when multiple terminal tasks finish at the same time, one with a failed state.

Fixes https://github.com/RackHD/RackHD/issues/295 along with https://github.com/RackHD/on-taskgraph/pull/137

@RackHD/corecommitters @stuart-stanley @VulpesArtificem

### DIFF
--- a/lib/workflow/stores/mongo.js
+++ b/lib/workflow/stores/mongo.js
@@ -585,20 +585,22 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
     };
 
     /**
-     * Checks whether there are any pending, reachable, tasks corresponding to
+     * Checks whether there are any pending, reachable, or failed tasks corresponding to
      * the given graph ID.
      * @param {Object} data
      * @param {String} data.graphId - The uniqe ID of the graph to be checked
      * @retuns {Promise} a promise for an object with a boolean 'done' field
      * @memberOf store
      */
-    exports.checkGraphFinished = function(data) {
+    exports.checkGraphSucceeded = function(data) {
         assert.object(data, 'data');
         assert.uuid(data.graphId, 'data.graphId');
 
         var query = {
             graphId: data.graphId,
-            state: Constants.Task.States.Pending,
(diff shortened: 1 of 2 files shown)


stuart-stanley:
Cool. One double-check question: is this going to have a problem with tasks that fail that have the allowedFail: true (or whatever that attribute is ;) ) set? I'm being lazy and not walking through the logic myself ;).

As long as that isn't a problem, +1.

benbp:
@stuart-stanley I think that is a valid issue with this. Also in most cases we will scrub older failed tasks from the database, but there could be possible cases where a failed, but completed task document gets pruned from the database right before the check for it, in which case we might see this race condition again (albeit much more rarely).

yyscamper:
jenkins: test this please

benbp:
@stuart-stanley ok, the potential "pruning" issue I brought up is irrelevant, because we only prune tasks from completed graphs.

In regards to the ignoreFailure issue, I'm going to address that as part of a fix for https://github.com/RackHD/RackHD/issues/327. I think we can possibly just add an additional task state called "ignoredFailure" or something.

benbp:
@yyscamper this will fail tests because it requires https://github.com/RackHD/on-taskgraph/pull/137 (they both require each other, unfortunately). I'll verify again that local integration test runs work and report back.

pscharla:
@RackHD/corecommitters Ben needs another review to get this one through.  Looks like the failing Jenkins test is only because of multiple PR dependencies.

stuart-stanley:
So this looks like it is all kind of intra-connected with https://github.com/RackHD/on-taskgraph/pull/137 .... I think this all comes together correctly, but there are a lot of moving parts coming in together. +1 on the general code, but please poke at it after all the PR's come in and make sure it doesn't go *boom* :)

PR #177, opened 2016-07-12
Add 'identified' and 'managed' node event

anhou:
For ORFS 246/247  docs: https://github.com/RackHD/docs/pull/296  'discovered' and 'removed' related PRs have been merged in previous PRs

## NOTE: WIP

@RackHD/corecommitters @RackHD/rackhd_dev

### DIFF
--- a/lib/models/node.js
+++ b/lib/models/node.js
@@ -209,16 +209,11 @@ function NodeModelFactory (
         publishEvent: function(node, next) {
             return waterline.obms.find({node: node.id})
             .then(function(obms) {
-                if(node.sku && _.isEmpty(obms)) {
+                if(node.sku) {
                     return eventsProtocol.publishNodeAlert(node.id,
                         { nodeId : node.id,
                           nodeType: node.type,
-                          state: "identified" });
-                } else if (node.sku && !_.isEmpty(obms)) {
-                    return eventsProtocol.publishNodeAlert(node.id,
-                        { nodeId : node.id,
-                          nodeType: node.type,
-                          state: "managed" });
+                          state: _.isEmpty(obms)? "identified":"managed" });
                 }
             }).then(function() {
                 next();
@@ -227,8 +222,10 @@ function NodeModelFactory (
             });
         },
 
-        afterUpdate: function (node, next) {
-            this.publishEvent(node, next);
+        beforeUpdate: function (node, next) {
+            if (node.sku) {
+                this.publishEvent(node, next);
+            }
         }
     });
 }
--- a/lib/models/node.js
+++ b/lib/models/node.js
@@ -12,11 +12,21 @@ NodeModelFactory.$inject = [
     'Promise',
     'Constants',
     'Services.Configuration',
+    'Protocol.Events',
     'Protocol.Waterline'
 ];
 
 var bson = require('bson');
-function NodeModelFactory (Model, waterline, _, Promise, Constants, configuration, waterlineProtocol) {
+function NodeModelFactory (
+    Model,
+    waterline,
+    _,
+    Promise,
+    Constants,
+    configuration,
+    eventsProtocol,
+    waterlineProtocol
+) {
     var dbType = configuration.get('databaseType', 'mongo');
 
     var attributes = {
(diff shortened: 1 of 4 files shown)


lib/models/node.js
```diff
@@ -194,6 +204,31 @@ function NodeModelFactory (Model, waterline, _, Promise, Constants, configuratio
                 return next();
             }
             return next();
+        },
+
+        publishEvent: function(node, next) {
+            return waterline.obms.find({node: node.id})
+            .then(function(obms) {
+                if(node.sku && _.isEmpty(obms)) {
```
WangWinson: Will below be a little easy ?

``` javascript
if (node.sku) {
    return eventsProtocol.publishNodeAlert(node.id, {
          nodeId : node.id,
          nodeType: node.type,
          state: _.isEmpty(obms) ? "identified" : "managed"
    });
}
```

yyscamper: Agree

lib/models/node.js
```diff
@@ -194,6 +204,31 @@ function NodeModelFactory (Model, waterline, _, Promise, Constants, configuratio
...
+                }
+            }).then(function() {
+                next();
+            }).catch(function(err) {
+                next(err);
+            });
+        },
+
+        afterUpdate: function (node, next) {
+            this.publishEvent(node, next);
```
yyscamper: What if only the node's name or identifier or other fields (these not either sku or obm) are changed? I think there is no need to publish event is such cases.

anhou: good catch, updated

lib/models/node.js
anhou: @WangWinson Good,  didn't pay attention, updated

lib/models/obms.js
```diff
@@ -106,6 +108,37 @@ function ObmsModelFactory (
...
+        publishEvent: function(nodeId, next) {
+            return waterline.nodes.findByIdentifier(nodeId)
+            .then(function(node) {
+                if (!_.isEmpty(node)) {
+                    return waterline.obms.find({node: node.id})
+                    .then(function(obms) {
+                        return eventsProtocol.publishNodeAlert(node.id,
+                            { nodeId : node.id,
+                              nodeType: node.type,
+                              state: _.isEmpty(obms) ? 'identified': 'managed' });
```
zyoung51: Do you want to check for the node.sku here to decide if it is 'identified'?

lib/models/obms.js
```diff
@@ -106,6 +108,37 @@ function ObmsModelFactory (
             next();
         },
 
+        publishEvent: function(nodeId, next) {
+            return waterline.nodes.findByIdentifier(nodeId)
```
zyoung51: Would 'needByIdentifier' be useful to clean up the logic below?

lib/models/obms.js
```diff
@@ -106,6 +108,37 @@ function ObmsModelFactory (
...
+            .then(function(node) {
+                if (!_.isEmpty(node)) {
+                    return waterline.obms.find({node: node.id})
```
zyoung51: If 'created/updated' is calling this can you skip going back to the database and assert that it is 'managed'.

zyoung51:
How does the addition of these AMQP notifications and extra database accesses affect the latency of database operations?  Can these notifications occur asynchronously without waiting for the promise to be resolved?  Is the inability to emit an event a good reason to error out the database operation?

lib/models/obms.js
anhou: WIP PR, many logic holes here

lib/models/obms.js
anhou: needByIdnetifier will throw Errors.NotFoundError,  not intended to let it throw error

anhou:
@zyoung51 Thanks for comments, it's a WIP PR to try to attempt to achieve publish node events in models,  the comments you address such as latency/performance, error handling is good concerns, and more, I found that some previous record information for stage change cannot be got from lifecycle functions, so I'll close this PR and create other PRs for 'identified' and 'managed'  events

lib/models/obms.js
zyoung51: Is your thought that the .catch() on line 125 would not be able to handle an exception thrown in the promise chain?

PR #185, opened 2016-07-21
Fix not returned promise

anhou:
Fix Jenkins failure, a mistake that missing return promise when no action for node's sku field. 
TypeError: Cannot read property 'catch' of undefined at /home/vagrant/src/on-tasks/lib/jobs/generate-sku.js:129

@RackHD/corecommitters @jlongever


jlongever:
+1

yyscamper:
+1

jlongever:
test this please

PR #186, opened 2016-07-26
Update JsonSchemaValidator to handle schema with namespace

WangWinson:
- Support customize namespace in `options.nameSpace` when instantiation.
- Add APIs `addSchema`, `addSchemas`, `addSchemasByDir` to add schemas via namespace.
- Add API `getAllSchemaNames` to return all schema names.
- Update API `getSchema` with namespace resolved.
- Add unit test for new added function.

@RackHD/corecommitters @iceiilin @cgx027 @pengz1

### DIFF
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -132,18 +132,20 @@ function jsonSchemaValidatorFactory(
      */
     JsonSchemaValidator.prototype.addSchemasByDir = function (schemaDir, metaSchemaName) {
         var self = this;
-
-        assert.string(schemaDir, 'schemaDir');
-        if (metaSchemaName) {
-            assert.string(metaSchemaName, 'metaSchemaName');
-        }
-
-        fs.readdirAsync(schemaDir)
+        return Promise.try(function () {
+            assert.string(schemaDir, 'schemaDir');
+            if (metaSchemaName) {
+                assert.string(metaSchemaName, 'metaSchemaName');
+            }
+        })
+        .then(function () {
+            return fs.readdirAsync(schemaDir);
+        })
         .filter(function (fileName) {
             return /\.json$/i.test(fileName);
         })
         .reduce(function (entries, fileName) {
-            return fs.readfileAsync(path.resolve(schemaDir, fileName))
+            return fs.readFileAsync(path.resolve(schemaDir, fileName))
             .then(function (fileData) {
                 var content = JSON.parse(fileData);
                 content.id = fileName;
@@ -156,7 +158,7 @@ function jsonSchemaValidatorFactory(
             if (metaSchemaName) {
                 var metaSchema = entries[metaSchemaName];
                 if (!metaSchema) {
-                    throw Errors.NotFoundError('Cannot find the meta schema "' + metaSchemaName +
+                    throw new Error('Cannot find the meta schema "' + metaSchemaName +
                         '" in directory "' + schemaDir + '"');
                 }
                 self.addMetaSchema(metaSchema);
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -8,20 +8,25 @@ jsonSchemaValidatorFactory.$provide = 'JsonSchemaValidator';
 jsonSchemaValidatorFactory.$inject = [
     'Ajv',
     'Assert',
+    'Errors',
     '_'
 ];
 
 function jsonSchemaValidatorFactory(
     Ajv,
     assert,
+    Errors,
     _
 ) {
     var url = require('url');
+    var fs = require('fs');
+    var path = require('path');
+    var defaultNameSpace = 'http://rackhd.com/schemas/v1/'; // TODO: need it or not
 
     function JsonSchemaValidator(options) {
         this.options = options || {};
+        this.nameSpace = options.nameSpace || defaultNameSpace;
         this._ajv = new Ajv(this.options);
-        this.addSchema = this._ajv.addSchema;
         this.addMetaSchema = this._ajv.addMetaSchema;
     }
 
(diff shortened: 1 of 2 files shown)


lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+     */
+    JsonSchemaValidator.prototype.addSchemasByDir = function (schemaDir, metaSchemaName) {
+        var self = this;
+
+        assert.string(schemaDir, 'schemaDir');
+        if (metaSchemaName) {
+            assert.string(metaSchemaName, 'metaSchemaName');
+        }
+
+        var fileNames = _.filter(fs.readdirSync(schemaDir), function (fileName) {
```
yyscamper: Change to async readdir

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+            var content = JSON.parse(data.toString());
+            content.id = name;
+            if (metaSchemaName === fileName) {
+                metaSchema = content;
+            } else {
+                schemas.push(content);
+            }
+        });
+
+        //meta schema should be added first, so subsequent schemas can be validated againt it
```
yyscamper: typo: againt -->against

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+        if (metaSchemaName) {
+            if (!metaSchema) {
+                throw Errors.NotFoundError('Cannot find the meta schema "' + metaSchemaName +
+                    '" in directory "' + schemaDir + '"');
+            }
+            self.addMetaSchema(metaSchema);
```
yyscamper: Move this line into the block of `if (!metaSchema)`

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+            if (metaSchemaName === fileName) {
+                metaSchema = content;
+            } else {
+                schemas.push(content);
+            }
+        });
+
+        //meta schema should be added first, so subsequent schemas can be validated againt it
+        if (metaSchemaName) {
+            if (!metaSchema) {
```
yyscamper: Suggest combining two `if` to `if (metaSchemaName && !metaSchema)`

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+            }
+            self.addMetaSchema(metaSchema);
+        }
+
+        self.addSchemas(schemas);
+    };
+
+    /**
+     * Get all schema names from the instance.
+     * @return {Array} schema names
```
yyscamper: Better to make the return type more clear: Array -->Array<String>

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+     */
+    JsonSchemaValidator.prototype.getAllSchemaNames = function() {
+        return _.filter(_.keys(this._ajv._schemas), function(schemaName) {
+            return !schemaName.startsWith('http://json-schema.org');
```
yyscamper: To filter out the meta schema as well?

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+        }
+
+        self.addSchemas(schemas);
+    };
+
+    /**
+     * Get all schema names from the instance.
+     * @return {Array} schema names
+     */
+    JsonSchemaValidator.prototype.getAllSchemaNames = function() {
```
yyscamper: Suggest adding a new argument to tell whether includes the namespace in returned schema names

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
...
+        var fileNames = _.filter(fs.readdirSync(schemaDir), function (fileName) {
+            return path.extname(fileName).toLowerCase() === '.json';
+        });
+
+        var schemas = [];
+        var metaSchema;
+        _.forEach(fileNames, function(fileName) {
+            var fullName = path.resolve(schemaDir, fileName);
+            var data = fs.readFileSync(fullName);
+            var name = path.basename(fileName.toLowerCase(), '.json');
```
yyscamper: For redfish schemas, there are both upper & lower cases for filename, I am afraid convert them to lowercase will fail to resolve some schema reference.

lib/common/json-schema-validator.js
```diff
@@ -90,25 +95,105 @@ function jsonSchemaValidatorFactory(
     };
 
     /**
-     * Get compiled schema from the instance by `key` or `ref`.
-     * @param  {String} key `key` that was passed to `addSchema` or
-     *                  full schema reference (`schema.id` or resolved id).
+     * Add schema to the instance.
+     * @param {Object} schema
+     * @param {String} name Optional schema name when schema has an id
```
yyscamper: To describe the optional argument, you can use
```
@param {String} [name] - Optional schema name
```
Reference: http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

lib/common/json-schema-validator.js
```diff
@@ -8,20 +8,25 @@ jsonSchemaValidatorFactory.$provide = 'JsonSchemaValidator';
...
 function jsonSchemaValidatorFactory(
     Ajv,
     assert,
+    Errors,
     _
 ) {
     var url = require('url');
+    var fs = require('fs');
+    var path = require('path');
+    var defaultNameSpace = 'http://rackhd.com/schemas/v1/'; // TODO: need it or not
```
yyscamper: rackhd.com is not our domain, I think this need be changed. Do you have an recommendation? @heckj @benbp @brianparry

lib/common/json-schema-validator.js
WangWinson: well,  I use `readdirSync` here , so that no need to using Promise

lib/common/json-schema-validator.js
WangWinson: I think the logic here is correct.
1. make sure `metaSchemaName` param passed and is not empty
2. if no `metaSchema` value found , throw error. 
3. otherwise , `addMetaSchema`

lib/common/json-schema-validator.js
WangWinson: Agree

lib/common/json-schema-validator.js
WangWinson: OK. better to use the original file name , not change the case

lib/common/json-schema-validator.js
WangWinson: Good stuff!

lib/common/json-schema-validator.js
WangWinson: Just a place holder here. Need to figure out a good way.

lib/common/json-schema-validator.js
yyscamper: It's my fault, you are correct.

lib/common/json-schema-validator.js
iceiilin: When will addSchemasByDir be used? Will sync dir reading impact system performance?

anhou: @WangWinson promise is not a problem, When having an I/O operation, it's better to use async method. otherwise, it will block  other codes, due to nodejs's single process.

WangWinson: The `addSchemasByDir` suppose to be used in `task-option-validator.register`, which will be called only once when taskgraph service start. 
Performance is a concern, but since this will be only run once at service start , so maybe we could ignore the impact.

yyscamper: Currently in production environment, we don't support runtime reboot our service, the `addSchemasByDir` is sure will only be called once during the service booting up. so I still think we should use async API:

- At least there will be some performance increase for service initialization time, especially when the schemas files are accumulated in a large number..
- The on-http unit-test will frequently start and stop service, so `addSchemasByDir` will be called frequently as well. Change to async will bring some performance increase for on-http unit-test. (On-http unit-test is very slow now, we'd try our best to not bring any more latency).

lib/common/json-schema-validator.js
```diff
@@ -90,25 +99,116 @@ function jsonSchemaValidatorFactory(
...
+     * Get all schema names from the instance.
+     * @param {Object} [option] - namespace will not be included by default,
+     *      set { includeNameSpace: true } when need to include namespace.
+     * @return {Array<String>} schemaNames
+     */
+    JsonSchemaValidator.prototype.getAllSchemaNames = function (option) {
+        var self = this;
+        return _(self._ajv._schemas).keys()
+        .filter(function (name) {
+            return name.startsWith(self.nameSpace);
```
yyscamper: If `nameSpace` is empty, then the `startsWith` will always true, but this is not expected:
```
> "abc".startsWith('')
true
```

WangWinson: Good Catch !

WangWinson:
Jenkins: test this please

WangWinson:
Jenkins: test this please

WangWinson:
Jenkins: test this please

yyscamper:
:+1:

cgx027:
+1

WangWinson:
Jenkins: test this please

yyscamper:
There is inter-dependency between this PR, the Jenkins failure is expected. So I need to force merge this PR. Any error will revert this change.

PR #187, opened 2016-07-29
Changes needed for docker tasks in on-tasks.

rolandpoulter:
* Adds ComputeContainer constant, and docker property to node model
* Adds custom timeout argument for messenger.subscribe

### DIFF
--- a/lib/common/messenger.js
+++ b/lib/common/messenger.js
@@ -255,9 +255,10 @@ function messengerFactory(
      * @param {string} routingKey
      * @param {Object} data
      * @param {Object} type The constructor of the expected type for the response.
+     * @param {number} Override default messenger timeout
      * @returns {Promise}
      */
-    Messenger.prototype.request = function (name, routingKey, data, type) {
+    Messenger.prototype.request = function (name, routingKey, data, type, timeout) {
         assert.string(name, 'name');
         assert.string(routingKey, 'routingKey');
         assert.object(data, 'data');
@@ -268,6 +269,7 @@ function messengerFactory(
         var context = {
             subscription: null,
             timeout: null,
+            timeoutDuration: timeout,
             name: name,
             routingKey: routingKey,
             type: type,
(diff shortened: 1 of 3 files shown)


rolandpoulter:
@RackHD/corecommitters please review this.

anhou:
I'm not sure if it's appropriate to add a non- bare metal node type here  'compute-container'  @RackHD/corecommitters  How do you think? or maybe need a discussion in CC meeting

rolandpoulter:
Thanks for the feedback @anhou 

I decided to add a `compute-container` node for two main reasons.

1) The nodes collection already has a nice method for storing relationships between nodes and I wanted there to be a one-to-many relationship between host nodes with docker and their containers.

2) I created a graph that will stop a `compute-container` and rather than using the host node as the target and specifying a container ID, it seemed be simpler to be able to target a node that represents the container you want to stop.

lib/common/messenger.js
```diff
@@ -255,9 +255,10 @@ function messengerFactory(
      * @param {string} routingKey
      * @param {Object} data
      * @param {Object} type The constructor of the expected type for the response.
+     * @param {number} Override default messenger timeout
      * @returns {Promise}
      */
-    Messenger.prototype.request = function (name, routingKey, data, type) {
+    Messenger.prototype.request = function (name, routingKey, data, type, timeout) {
```
benbp: yay

rolandpoulter:
We could add a boolean field to the nodes collection to specify whether or not the node is a physical or virtual resource? That way you could easily filter the nodes collection to hardware only.

benbp:
@anhou @rolandpoulter I don't think there's any harm in adding some low-usage, experimental node types and seeing what people decide to do with them (kind of as an experiment, I guess). Might warrant a discussion though, yeah.

rolandpoulter:
@anhou @heckj @yyscamper @tldavies @brianparry @benbp @pscharla 
Anyone have any concerns or thoughts on adding `compute-container` as a non-bare metal node type for docker linux containers?

heckj:
@rolandpoulter we talked about it earlier today (https://groups.google.com/forum/#!topic/rackhd/rEQ-a_frQBs) - good by me for inclusion

heckj:
👍

brianparry:
:+1:

PR #189, opened 2016-08-08
add schemaRef in task-definition model

yyscamper:
To support Task-Schema, a new key "schemaRef" is defined to store the reference of its task schema. Now it's set to be optional as not all existing tasks have associated schemas.

@RackHD/corecommitters @WangWinson @iceiilin

### DIFF
--- a/spec/lib/models/task-definition-spec.js
+++ b/spec/lib/models/task-definition-spec.js
@@ -105,6 +105,20 @@ describe('Models.Task', function () {
                 expect(this.subject.json).to.equal(true);
             });
         });
+
+        describe('schemaRef', function () {
+            before(function () {
+                this.subject = this.attributes.schemaRef;
+            });
+
+            it('should be optional', function () {
+                expect(!this.subject.required).to.equal(true);
+            });
+
+            it('should be string', function () {
+                expect(this.subject.type).to.equal('string');
+            });
+        });
     });
 
     describe('object returned from toJSON()', function () {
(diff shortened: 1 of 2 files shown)


spec/lib/models/task-definition-spec.js
```diff
@@ -105,6 +105,20 @@ describe('Models.Task', function () {
...
             });
         });
+
+        describe('schemaRef', function () {
+            before(function () {
+                this.subject = this.attributes.schemaRef;
+            });
+
+            it('should be optional', function () {
+                expect(!this.subject.required).to.equal(true);
```
WangWinson: why not ?
``` Javascript
expect(this.subject.required).to.not.equal(true)
```

WangWinson:
:+1:

spec/lib/models/task-definition-spec.js
yyscamper: I want to test this is not a truthy value, not the actual `true`. The `equal` will not work well as my expectation, I finally found the `ok` is what I want:
```javascript
expect('everything').to.be.ok;
expect(1).to.be.ok;
expect(false).to.not.be.ok;
expect(undefined).to.not.be.ok;
expect(null).to.not.be.ok;
```

iceiilin:
:+1:

yyscamper:
jenkins: test this please

anhou:
:+1:

yyscamper:
jenkins: test this please

PR #190, opened 2016-08-08
Redact passwords

DavidjohnBlodgett:
resolves RackHD/RackHD#241

@RackHD/corecommitters

new PR at the request of @brianparry 's to place logic into graph-object model.

due to a number of issues hit attempting to make my earlier code testable, I refactored to use encryption so passwords are not saved in our database in plain text to begin with and will be naturally concealed from the user on a query.

unit tests are complete for this feature.

### DIFF
--- a/lib/common/sanitizer.js
+++ b/lib/common/sanitizer.js
@@ -1,0 +1,52 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+module.exports = SanitizerFactory;
+
+SanitizerFactory.$provide = 'Sanitizer';
+SanitizerFactory.$inject = [
+    'Services.Encryption',
+    'Constants',
+    'Util',
+    '_'
+];
+
+function SanitizerFactory (
+    encryption,
+    Constants,
+    util,
+    _
+) {
+
+    var secretsPattern = _.reduce(Constants.Logging.Redactions, function(pattern, regex) {
+        return pattern ? util.format('%s|(?:%s)', pattern, regex.source) :
+                         util.format('(?:%s)', regex.source);
+    }, '');
+    var secretsRegex = new RegExp(secretsPattern, 'i');
+
+    function sanitizer (obj) {
+        _.forOwn(obj, function(value, key) {
+            if (typeof(value) === 'object') {
+                sanitizer(value);
+            } else if ( key.match( secretsRegex )) {
+                obj[key] = encryption.encrypt(value);
+            }
+        });
+    }
+
+    function de_sanitizer (obj) {
+        _.forOwn(obj, function(value, key) {
+            if (typeof(value) === 'object') {
+                de_sanitizer(value);
+            } else if ( key.match( secretsRegex )) {
+                obj[key] = encryption.decrypt(value);
+            }
+        });
+    }
+
+    return {
+        encrypt: sanitizer,
+        decrypt: de_sanitizer
+    }
+}
--- a/lib/workflow/stores/mongo.js
+++ b/lib/workflow/stores/mongo.js
@@ -9,10 +9,11 @@ mongoStoreFactory.$inject = [
     'Constants',
     'Errors',
     'Assert',
-    '_'
+    '_',
+    'Sanitizer'
 ];
 
-function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
+function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _, sanitizer) {
     var exports = {};
 
     exports.publishRecordByGraphId = function (graphId, event) {
@@ -202,7 +203,6 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             new: true,
             upsert: true
         };
-
         return waterline.graphdefinitions.findAndModifyMongo(query, {}, definition, options);
     };
 
@@ -240,7 +240,6 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             new: true,
             upsert: true
         };
-
         return waterline.taskdefinitions.findAndModifyMongo(query, {}, definition, options);
     };
 
@@ -307,7 +306,6 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
                 instanceId: 1
             }
         };
-
         return waterline.graphobjects.findAndModifyMongo(query, {}, graph, options);
     };
 
@@ -370,6 +368,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
 
         return waterline.graphobjects.findOne(query, options)
         .then(function(graph) {
+            sanitizer.decrypt(graph.tasks[data.taskId])
             return {
                 graphId: graph.instanceId,
                 context: graph.context,
--- a/lib/common/model.js
+++ b/lib/common/model.js
@@ -14,7 +14,8 @@ ModelFactory.$inject = [
     'Errors',
     'Waterline',
     'Protocol.Waterline',
-    'Services.Waterline'
+    'Services.Waterline',
+    'Sanitizer'
 ];
 
 function ModelFactory (
@@ -26,7 +27,8 @@ function ModelFactory (
     Errors,
     Waterline,
     waterlineProtocol,
-    waterline
+    waterline,
+    sanitizer
 ) {
     var url = require('url');
 
@@ -195,6 +197,9 @@ function ModelFactory (
             }
             var update = args.shift();
 
+            // encrypt password fields
+            sanitizer.encrypt(update);
+
             // Don't validate if we're setting/unsetting individual properties
             if (update.$set || update.$setOnInsert) {
                 if (update.$setOnInsert) {
@@ -223,6 +228,7 @@ function ModelFactory (
             .then(function() {
                 return self.runNativeMongo('findAndModify', _arguments);
             });
+
         },
 
         updateMongo: function(filter, update, options) {
--- a/spec/lib/common/sanitizer-spec.js
+++ b/spec/lib/common/sanitizer-spec.js
@@ -1,0 +1,48 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+describe('Sanitizer', function () {
+    var graphObj, sanitizer;
+    helper.before();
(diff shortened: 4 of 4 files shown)


lib/models/graph-object.js
```diff
@@ -43,6 +43,38 @@ function GraphModelFactory (Model, Constants, configuration) {
...
+                                        break;
+                                    }
+                                }
+                                if ( !alreadyFound ) {
+                                    walked.push( workflow[ property ]);
+                                    stack.push( { workflow : workflow[ property ], stack: item.stack + '.' + property });
+                                }
+                            } else {
+                                if ( property.match( /password/i )) {
+                                    workflow[property] = '*** REDACTED ***';
```
yyscamper: I hope we can use the same redacted word, see: https://github.com/RackHD/on-core/blob/master/lib/serializables/log-event.js#L193

lib/models/graph-object.js
```diff
@@ -43,6 +43,38 @@ function GraphModelFactory (Model, Constants, configuration) {
             active: function() {
                 var obj = this.toObject();
                 return Constants.Task.ActiveStates.indexOf(obj._status) > -1;
+            },
+            toJSON: function () {
```
yyscamper: I don't see the `toJSON` is called in API, so you need submit another PR to make sure the `toJSON` is called.
- https://github.com/RackHD/on-http/blob/a2da546c01dd7176a01f6bc93e8162e5bd1797e8/lib/api/1.1/northbound/workflows.js#L147
- https://github.com/RackHD/on-http/blob/a2da546c01dd7176a01f6bc93e8162e5bd1797e8/lib/api/2.0/workflows.js#L59

DavidjohnBlodgett: @yyscamper so Felix, the toJSON is an override of the base function in waterline for toJSON, so this is actually invoked automatically when a query is made (i've tested this)... however... I cannot have my cake and eat it too... I am unable to make unit tests using an override function like this, because of scope issues... I'll have to revert to making a class function and calling it in on-http like you suggest.

lib/models/graph-object.js
DavidjohnBlodgett: @yyscamper ok, so you would prefer "[REDACTED]" vs "*** REDACTED ***" ? just making sure I understand the request.

spec/lib/models/graph-object-spec.js
```diff
@@ -5,6 +5,7 @@
 
 var base = require('./base-spec');
 var  sandbox = sinon.sandbox.create();
+var graphObj;
```
DavidjohnBlodgett: oops missed removing this...

DavidjohnBlodgett:
so the most recent changes, we have gone with encrypting the password property, this way we can keep it encrypted on disk when a graph is shipped to a collection and not have an expensive recursive search done with each api call/query. new output looks like this: ```                "repo": "{{api.server}}/esxi/{{options.version}}",
                "rootPassword": "ezgZbdAHMhU9+KViV5A4nA==.hSP54vlC8tvWWVhS37yb9m8elNOW2TLIeMyDpSGrZJU=",
                "users": [
                    {
                        "name": "onrack",
                        "password": "FDWcWfw054YE9qWFmE87Ug==.6X+GyHREqap6VsSE9iuAi+BDXb6BWCc/VhImrgztwWQ=",
                        "uid": 1010
                    }
                ],```

DavidjohnBlodgett:
@brianparry so, I was wondering if I could get a second set of eye's on this (with ben being on vaca) so I'm worried, since this now encrypts passwords in the database... could other areas of our code expect plain text? and if that was the case... do these errors in jenkins seem to align with that thought process? thanks for any insight you can give!

amymullins:
test this please

lib/models/graph-object.js
yyscamper: no preference for me, just want to keep them using same value

lib/common/sanitizer.js
```diff
@@ -0,0 +1,57 @@
...
+     */
+    return function Sanitizer (workflow) {
+        //workflow = JSON.parse(workflow);
+        var sanitizedworkflow = workflow;
+        var walked = [];
+        var stack = [ { workflow : workflow , stack : '' } ];
+        while ( stack.length > 0 ) {
+            var item = stack.pop();
+            var workflow  = item.workflow;
+            for ( var property in workflow ) {
```
brianparry: Minor comment: lodash `forOwn` could simplify this logic a bit.

lib/common/sanitizer.js
```diff
@@ -0,0 +1,57 @@
...
+                                alreadyFound = true;
+                                break;
+                            }
+                        }
+                        if ( !alreadyFound ) {
+                            walked.push( workflow[ property ]);
+                            stack.push( { workflow : workflow[ property ], stack: item.stack + '.' + property });
+                        }
+                    } else {
+                        if ( property.match( /password/i ) || property.match(/community/i)) {
```
brianparry: There is a constant array of redaction patterns you can use for this. See https://github.com/RackHD/on-core/blob/master/lib/models/obms.js#L36

lib/common/sanitizer.js
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     *
+     */
+    return function Sanitizer (workflow) {
+        //workflow = JSON.parse(workflow);
+        var sanitizedworkflow = workflow;
+        var walked = [];
+        var stack = [ { workflow : workflow , stack : '' } ];
+        while ( stack.length > 0 ) {
```
brianparry: General question on this function: rather than pushing object properties onto a stack, would it be easier to use a recursive function?  Maybe something like: 
```
function encryptObj(obj) {
    _.forOwn(obj, function(value, key) {
        if (typeof(value) === 'object') {
            encryptObj(value);
        } else if (key.match(secretsPattern) {
            obj[key] = encrypt(value);
        }
    }
}
```

lib/workflow/stores/mongo.js
```diff
@@ -202,7 +203,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             new: true,
             upsert: true
         };
-
+        definition = sanitizer(definition);
```
brianparry: @DavidjohnBlodgett Could you move the sanitizer call into `findAndModifyMongo`?

brianparry:
Thanks @DavidjohnBlodgett!  :+1: aside from the unit test failure.

yyscamper:
Two questions:
- Since you encrypt the options while persisting graph object, so how and when to decrypt the password before if the job wants to get the real value?
- Output the ciphertext in API response still have security concern, since RackHD is open source, so it's clear that how to decrypt it. So this link redaction still leak the sensitive information.

brianparry:
@yyscamper For obms, I added some convenience methods in the model to decrypt, but ultimately this should be the responsibility of the job that uses the secret.
On your second point, the problem is not that we're revealing the cipher; it's that we're storing our keys insecurely.  Really, we should be using something like this: https://www.hashicorp.com/blog/vault.html

DavidjohnBlodgett:
@yyscamper hey Felix, so I believe your first question is the root problem for why some of the integration tests are currently failing... we are not decrypting the password before it is being used.... so yes, this is an open question, I am assuming I'll have to hunt down the locations we need to decrypt and add those to the PR. For the short term I would agree with @brianparry to have the features that use passwords be expected to decrypt.

DavidjohnBlodgett:
@yyscamper @brianparry I am open to any suggestions (and ripping this logic out if we add database tagging), and can look into the vault option as a perhaps long term solution. (at this time, I was focused on a short term fix for some blocking ODR stuff and this seemed like a more elegant method of masking our sensitive data then just transforming an object from a database that's storing plaintext ) I do however understand that it isn't entirely secure... just better then what we got lol

DavidjohnBlodgett:
@brianparry just a heads up, I implemented most of what you asked for... still looking at shifting up the sanitizer function call into findAndModifyMongo~ (having a nubby moment). In addition, maybe seeking some help sorting out what jobs exactly are failing that could lead to where we expect the passwords to be decrypted...

pscharla:
test this please

DavidjohnBlodgett:
I think I've tracked down this jenkins failure to 
```
Task.Base.Linux.Commands
```

working on a fix

tldavies:
test this please

lib/workflow/stores/mongo.js
```diff
@@ -202,7 +203,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             new: true,
             upsert: true
         };
-
+        //sanitizer(definition);
```
yyscamper: Remove this line if you don't want it.

yyscamper:
@DavidjohnBlodgett : 
The biggest problem of the PR is that it lacks the peer function of `santize`, the data is encrypted but it is not decrypted in job, so that if your PR is merged, I can think of there is a lot jobs will not work well, such as install-os.

DavidjohnBlodgett:
@yyscamper I cannot agree more, I've been trying to wrap my brain around all possible effected areas. If you can think of any other area's we may need to unencrypt aside from OS installs and maybe ssh command tasks... That would be very helpful... I'm assuming I'll have to post a dependent PR in on-tasks to implement to decrypting stage for each specific job... unless there are any possible layers of abstraction I am unaware of that decryption could reside. There's a part of me that feels we may have bitten off more then we can chew, and may need to revert back to simple redaction at the api layer if this is a blocking issue for any internal group...

DavidjohnBlodgett:
please don't merge yet... manual testing has found this new logic seems to miss a rootPassword property in my example, thats found outside of a workflows options... investigating now...

DavidjohnBlodgett:
@RackHD/corecommitters ok guys... I think this is it... we made some changes to insure that decryption happens when we actually use it by tasks... and I manually tested that the recursion and regex work. This should be merge ready. Let me know if you have any questions, or concerns!

rolandpoulter:
👍 although you could squash the commit history down a little.

DavidjohnBlodgett:
@rolandpoulter understood.

pscharla:
@RackHD/corecommitters Need merge for this one to resolve the P1 ODR for upcoming release.

brianparry:
Looks good to me :+1:

PR #192, opened 2016-08-10
Add revealSecrets option to models.obms.upsertByNode

benbp:
Adding the option to return decrypted values when calling the obm model's upsertByNode method. Supports https://github.com/RackHD/on-tasks/pull/313

@RackHD/corecommitters @hohene @johren


anhou:
:+1:

yyscamper:
:+1:

PR #202, opened 2016-08-31
added findAllByNode to Obms

geoff-reid:
This is to support the GET route for nodes/id/obm
@brianparry @RackHD/corecommitters @BillyAbildgaard @srinia6 @dalebremner @keedya

### DIFF
--- a/lib/models/obms.js
+++ b/lib/models/obms.js
@@ -208,6 +208,22 @@ function ObmsModelFactory (
             });
         },
 
+        findAllByNode: function(nodeId, reveal, query) {
+            var self = this;
+
+            query = query || {};
+            return Promise.try(function() {
+                    _.merge(query, {node: nodeId} );
+                    return self.find(query);
+                })
+                .then(function(obms) {
+                    if (obms && reveal) {
+                        return obms.map(_revealSecrets);
+                    }
+                    return obms;
+                });
+        },
+
         secret: function(attr) {
             return attr.match(secretsPattern);
         }
--- a/spec/lib/models/obms-spec.js
+++ b/spec/lib/models/obms-spec.js
@@ -215,6 +215,27 @@ describe('Models.Obms', function () {
                 expect(obm).to.be.undefined;
             });
         });
+
+        it('should find all nodes for nodeId', function() {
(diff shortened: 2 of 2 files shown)


lib/models/obms.js
```diff
@@ -208,6 +208,22 @@ function ObmsModelFactory (
...
+            var self = this;
+
+            query = query || {};
+            return Promise.try(function() {
+                    _.merge(query, {node: nodeId} );
+                    return self.find(query);
+                })
+                .then(function(obms) {
+                    if (obms && reveal) {
+                        return obms.map(_revealSecrests);
```
yyscamper: The function name is not correct, it is `_revealSecrets` rather than your `_revealSecrests`

yyscamper:
Please add the unit-testing for the new function.

brianparry:
:+1:

spec/lib/models/obms-spec.js
```diff
@@ -215,6 +215,27 @@ describe('Models.Obms', function () {
...
+                    expect(obm.toJSON().config).not.to.have.property('password');
+                });
+            });
+        });
+
+        it('should find all nodes for nodeId and reveal passwords', function() {
+            return obms.findAllByNode(nodeId, true, {service: 'ipmi-obm-service'})
+            .then(function(obms) {
+                expect(obms).to.have.length(1);
+                expect(obms[0].node).to.equal(nodeId);
```
yyscamper: The last test case checks all obms, why this test case only check the first obm? Check all obms will verify the _revealSecrets works for all obms.

yyscamper:
:+1: after changing unit-testing to verify all obms.

keedya:
looks good to me :+1:

spec/lib/models/obms-spec.js
geoff-reid: The previous test tests for all obms, I was using this test to test the query function.  Also, only the first example OBM has a password, so the _revealSecrets will only work for it.

yyscamper: Actually, the `community` is also a kind of sensitive data, it will also be encrypted. so that's OK to check all obms.

PR #204, opened 2016-09-01
Optimize workflow mongo store publishRecordByGraphId to find/send les…

benbp:
…s data

If we have a bunch of tasks being updated at one time (at least dozens per second), then grabbing the entire graph object out of mongo and sending it over AMQP can add some unnecessary CPU load. This modifies the code to just grab the graph instanceId, status, and/or error fields instead. Will require some corresponding changes to the UI operations center to do a follow-up query on these events I believe (CC @rolandpoulter).

Doing some really, really rough benchmarking (a.k.a. eyeballing htop), running this function every 25ms had about 50% the CPU load as the old way (5-7% vs. 11-15%).

@RackHD/corecommitters @zyoung51 @jlongever @VulpesArtificem @pscharla

### DIFF
--- a/lib/common/model.js
+++ b/lib/common/model.js
@@ -182,8 +182,8 @@ function ModelFactory (
             });
         },
 
-        findOneMongo: function(query) {
-            return this.runNativeMongo('findOne', [query]);
+        findOneMongo: function(args) {
+            return this.runNativeMongo('findOne', args);
         },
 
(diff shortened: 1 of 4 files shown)


yyscamper:
:+1:

zyoung51:
+1

rolandpoulter:
:+1:

PR #205, opened 2016-09-02
Issue#385: Fix unit-test error due to model.findSinceLastUpdate

yyscamper:
resolves https://github.com/RackHD/RackHD/issues/385

The function `findSinceLastUpdate` lookups by `updatedAt`, but the unit-testing is using `createdAt`. Since the all time values include `updatedAt`, `createdAt` for all records are very close, so the time value in unit-testing doesn't always work as expected.

I have tested the unit-testing for more than 100 times, no error happens again.

@DavidjohnBlodgett @rolandpoulter @RackHD/corecommitters

### DIFF
--- a/spec/lib/common/model-spec.js
+++ b/spec/lib/common/model-spec.js
@@ -447,6 +447,7 @@ describe('Model', function () {
 
     describe('multiple records', function () {
         var records = [];
+        var beforeTime, midTime, afterTime;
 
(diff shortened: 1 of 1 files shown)


DavidjohnBlodgett:
:+1: awesome!

rolandpoulter:
👍

brianparry:
:+1:

PR #206, opened 2016-09-02
issue#326: change the log level of using default configuration value from info to warning

yyscamper:
resolves https://github.com/RackHD/RackHD/issues/326

@heckj @RackHD/corecommitters

### DIFF
--- a/lib/services/configuration.js
+++ b/lib/services/configuration.js
@@ -70,7 +70,7 @@ function configurationServiceFactory(
         var value = nconf.get(key);
 
         if (value === undefined) {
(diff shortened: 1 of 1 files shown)


heckj:
I'm totally good with this change - but is this warning even of value? Do we care when we a default value vs. a value set in a configuration file?

👍

anhou:
:+1:

yyscamper:
@heckj : your concern makes sense, I agree the `warning` will be better than `debug`. The PR has been updated with your comment is addressed.

cgx027:
:+1:

PR #207, opened 2016-09-06
Index graph object model instanceId attribute

benbp:
Given this is our primary lookup field, we should probably index it.

@RackHD/corecommitters @VulpesArtificem

### DIFF
--- a/lib/models/graph-object.js
+++ b/lib/models/graph-object.js
@@ -20,7 +20,8 @@ function GraphModelFactory (Model, Constants, configuration) {
                 type: 'string',
                 required: true,
                 unique: true,
-                uuidv4: true
+                uuidv4: true,
+                index: true
             },
             context: {
                 type: 'json',


heckj:
:+1:

heckj:
:+1:

VulpesArtificem:
👍

PengTian0:
test this please

panpan0000:
test this please

panpan0000:
@benbp @yyscamper 
Jenkins encounter memory leak , so GC overhead limit exceeded ,and most job fails .
we re-start jenkins and re-test those PRs.

brianparry:
@benbp I'm not sure this change is necessary.  If the unique property is set for a model attribute, it should automatically be indexed by sails-mongo: https://github.com/balderdashy/sails-mongo/blob/62a3ca0a1d6e3088f420c1a606b643a6dae1296d/lib/collection.js#L407.  Also we probably want to be setting the `primaryKey` property.

brianparry:
@benbp Actually, I don't think this works at all if we have our default migration strategy set to `safe`.  See https://github.com/balderdashy/sails-mongo/issues/290.  I think we need to manually create indexes by adding a `setIndexes class method to the graph object model similar to what is done for lookups and obms.

benbp:
@brianparry ah! Ok, well in that case I'll re-open https://github.com/RackHD/on-taskgraph/pull/167, if you don't mind reviewing.

PR #212, opened 2016-09-20
add scope as a field when finding file in renderable db

iceiilin:
Currently, when finding a template, all files with the same name will be returned, despite of the desired scope. If the there are 3 scopes (say ["a", "b", "global"]) in database, but user only want to find two of them (["a", "global"]), it is possible that the final file would be for scope "b" after sorting by priority.

Add scope filed when querying database.

@zyoung51

### DIFF
--- a/lib/common/db-renderable-content.js
+++ b/lib/common/db-renderable-content.js
@@ -191,7 +191,7 @@ function dbRenderableContentServiceFactory(
             i += 1;
         });
 
-        return waterline[self.collectionName].find({name: name})
+        return waterline[self.collectionName].find({name: name, scope: scope})
             .then(function(templates) {
                 if (!templates.length) { return undefined; }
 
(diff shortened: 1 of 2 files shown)


lib/common/db-renderable-content.js
```diff
@@ -191,7 +191,7 @@ function dbRenderableContentServiceFactory(
             i += 1;
         });
 
-        return waterline[self.collectionName].find({name: name})
+        return waterline[self.collectionName].find({name: name, scope: scope})
```
zyoung51: Good catch.  The query was structured this way because the Cassandra waterline adapter does not support compound keys and will not support the IN query on the scope key due to that restriction.

iceiilin: Since Cassandra adapter doesn't support combined search, I changed to use a filter for designated scope after waterline.find(name) returns. @zyoung51

rackhdmirror:
Build finished.

iceiilin:
test this please

rackhdmirror:
Build finished.

yyscamper:
good finding.
:+1:

PR #213, opened 2016-09-22
actually ignore failures if so marked

VulpesArtificem:
fix for https://github.com/RackHD/RackHD/issues/327 

goes with 
https://github.com/RackHD/on-tasks/pull/344
and 
https://github.com/RackHD/on-taskgraph/pull/172

@RackHD/corecommitters @pscharla

### DIFF
--- a/lib/workflow/stores/mongo.js
+++ b/lib/workflow/stores/mongo.js
@@ -333,7 +333,8 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _, san
             graphId: graphId,
             state: Constants.Task.States.Pending,
             dependencies: taskDependencyItem.dependencies,
-            terminalOnStates: taskDependencyItem.terminalOnStates
+            terminalOnStates: taskDependencyItem.terminalOnStates,
+            ignoreFailure: taskDependencyItem.ignoreFailure
         };
 
         return waterline.taskdependencies.create(obj);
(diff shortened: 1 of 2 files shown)


lib/workflow/stores/mongo.js
```diff
@@ -621,6 +622,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _, san
             state: {
                 $ne: Constants.Task.States.Succeeded
             },
+            ignoreFailure: false,
```
benbp: Is it possible for this field to be `null` instead of `false` if not specified, preventing graphs without ignoreFailure tasks from succeeding?

benbp: Or non-existant?

VulpesArtificem: whoops, I meant {$ne: true } will change

rackhdmirror:
Build finished.

pscharla:
@RackHD/corecommitters: please review in the near term. This PR is blocking a couple P1 ODRs

pscharla:
@RackHD/corecommitters : Checking with @panpan0000 again, but last I knew we can ignore the "Jenkins-Dependency-PR" failure. Please merge since this is blocking P1 ODR.

johren:
Can this go in without https://github.com/RackHD/on-taskgraph/pull/172 or do we need +1s on that first?

pscharla:
@johren Looks like https://github.com/RackHD/on-taskgraph/pull/172 has 2 +1's. And I think that one needs this PR and https://github.com/RackHD/on-tasks/pull/344 to be merged first.

johren:
Yeah, I noticed that I had a stale copy of  RackHD/on-taskgraph#172 in my browser right after I added those notes so I deleted them.  You're too quick and  caught it before I deleted.  😜

PR #214, opened 2016-09-29
Updated node events to include additional data when published

dmskinner:
Related changes in `on-tasks`
https://github.com/RackHD/on-tasks/pull/350

Publish node event message with additional data, if provided.


rackhdmirror:

PR #217, opened 2016-10-19
support task-schema refactor

yyscamper:
Change including:
(1) Task-Definition: replace schemaRef with optionsSchema, optionsSchema can be either string or object, so set its type to `json`. 
(2) add removeSchema API for JSON schema validator;

@RackHD/corecommitters @iceiilin @pengz1 @cgx027 @lanchongyizu @amymullins

### DIFF
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -31,6 +31,7 @@ function jsonSchemaValidatorFactory(
         this.nameSpace = options.nameSpace || '';
         this._ajv = new Ajv(this.options);
         this.addMetaSchema = this._ajv.addMetaSchema;
+        this.removeSchema = this._ajv.removeSchema;
     }
 
     /**
--- a/lib/models/task-definition.js
+++ b/lib/models/task-definition.js
@@ -38,8 +38,8 @@ function TaskModelFactory (Model, configuration) {
                 required: true,
                 json: true
             },
-            schemaRef: {
-                type: 'string'
+            optionsSchema: {
+                type: 'json'
             },
             toJSON: function() {
                 // Remove waterline keys that we don't want in our graph objects
--- a/spec/lib/models/task-definition-spec.js
+++ b/spec/lib/models/task-definition-spec.js
@@ -106,17 +106,17 @@ describe('Models.Task', function () {
             });
         });
 
-        describe('schemaRef', function () {
+        describe('optionsSchema', function () {
             before(function () {
-                this.subject = this.attributes.schemaRef;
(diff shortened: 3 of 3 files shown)


rackhdmirror:
Build finished.

brianparry:
Hi @yyscamper.  I'm curious where this PR stands at the moment.  Do you intend to make any additional changes such as those we discussed during your visit?

yyscamper:
@brianparry No additional change, I was trying to resolve the conflict and rebase my code to make sure the  dependency Jenkins passes, then I will base on this code to do stress testing. This is the reason why you see the Jenkins becomes active again.

cgx027:
Looks good to me. :+1:

PengTian0:
test this please

PengTian0:
test this please

PengTian0:
test this please

changev:
test this please

changev:
test this please

PengTian0:
test this please

changev:
test this please

brianparry:
@yyscamper The change looks good, so :+1: , but I think we should consider adding parameter constraints inline with task definition properties rather than defining a seperate schema property.  One way to think about this is that the task definitions themselves would be a kind of schema.  I think this would make writing task definitions easier for users.

yyscamper:
@brianparry : Thanks for your review. But I don't quite understand your comment, whether you mean the schema should go along with each option rather than put together in a separate property? Something like below? 
```javascript
{
    injectableName: 'Task.InstallCentOS',
    friendlyName: 'CentOS install',
    options: {
        version: {
            type: 'string',
            required: true
        },
        repo: {
            type: 'string',
            required: true,
            default: '{{api.server}}/centos/{{options.version}}'
        }
    }
}
``` 

I did think about this, the challenge is it cannot or hard meet some complex case, for example, for LinuxCommand task, the `command` can either be a string or an object, so it should look like
```javascript
options: {
  command: {
     oneOf:[
        { 
           type: 'string'
        },
        { 
           type: 'object',
           properties: {
              ...
           }
        }
     ]
  }
}
```
It just like one task definition contains multiple sub json schema. For my case, all sub json schemas are contained in the seperated key `optionsSchema`.

Another reason is for some complex `required` case, for example, the CreateMegaRaid task can either be choosed to create default RAID or create user customized RAID, the `required` looks like:
```javascript
oneOf: [
    { 'required': ['controllerId', 'createDefault'] }
    { 'required': ['controllerId', 'raidConfig'] }
]
```
This kind of constraint applies for the whole of options rather than a single option.

iceiilin:
+1

PR #218, opened 2016-10-19
Ssh obm service

srinia6:
Create inband-mgmt model. This PR includes:
- Move sshSettings out of nodes.
- Move the older obms model code to a base called as network-management.
- create ibms collection.

@RackHD/corecommitters @brianparry @dalebremner @geoff-reid @BillyAbildgaard


rackhdmirror:

benbp:
Not sure how to reconcile what I see as a semantic issue (or whether to accept it and ignore it). I think that logically the OBM model/structure makes sense for leveraging SSH operations, but I would say that SSH isn't actually "out of band" in the way IPMI/other services are, because it's not usually on a segregated network.

srinia6:
@benbp yes, ssh settings now are part of in-band [ibms].  The obm and in-band (ibms ) now are 2 seperate models that use the common code (earlier know as obms.js) . Adding @brianparry to keep me honest here.

yyscamper:
@srinia6 : I have same concern with @benbp , the ibm is almost same with obm, but we created a lot of duplicated code to support it.

I think the biggest problem is the name `obm` itself, if at beginning, the name is not `obm` but rather something like "management interface", then the SSH can be easily added into. 

Comparing with adding so many duplicated code, I prefer to rename the obm, if we want to distinguish `ibm` and `obm`, we can add a new flag in renamed resource.

I wonder whether you have other particular reason for your proposed "ibm" that cannot be well fitted into "obm" so that we have to created "ibm"?

brianparry:
@benbp @yyscamper I'm a little confused.  I agree with your comments that ssh settings (and snmp) have different semantics than OBMs, but that's really the whole reason for this change.  @srinia6 extracted most of the OBMs code into a "base" model called `NetworkManagementModel`.  From this, she derives two other models `InBandMgmtModel` and `ObmModel`.  These models map to the `ibms` and `obms` collections respectively.  This retains the structure of OBMs while creating a new collection for in-band management settings.  The base model approach was used to avoid duplicated code.  This convention is also used for renderables such as templates, profiles, and views.  I think there is still some cleanup that needs to happen in the base model because it references OBMs in many places, but I think the overall structure of this change makes sense.

srinia6:
@brianparry - Thanks! Changed the obm variable name under NetworkManagementModel to a generic one.   @yyscamper - I am completely lost with your comments.  SSH is not part of obm's collection.  And the actual changes does eliminate code duplication.

srinia6:
Jenkins: depends on https://github.com/RackHD/on-tasks/pull/361

srinia6:
test this please

srinia6:
@benbp @yyscamper @RackHD/corecommitters - Somehow my PR shows previous 2 PR's as part of this one. To avoid confusion, I created another PR that diff's only my changes. 
Please refer - https://github.com/RackHD/on-core/pull/222

srinia6:
Opened new PR - https://github.com/RackHD/on-core/pull/222

benbp:
Yep, apologies @srinia6 . I was primed a little by the "obm" in the PR title, so clearly I misread the code.

PR #220, opened 2016-10-20
Fixed decryption when rendering a template file

DavidjohnBlodgett:
resolves RackHD/RackHD#417

@RackHD/corecommitters

@keedya

Currently fixes getting encrypted passwords from a template file. 

This and https://github.com/RackHD/on-tasks/pull/363 fixed issues with our JSON validator.

### DIFF
--- a/lib/common/json-schema-validator.js
+++ b/lib/common/json-schema-validator.js
@@ -12,7 +12,8 @@ jsonSchemaValidatorFactory.$inject = [
     'path',
     'Promise',
     'url',
-    '_'
+    '_',
+    'Sanitizer'
 ];
 
 function jsonSchemaValidatorFactory(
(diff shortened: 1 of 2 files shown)


rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

lib/common/json-schema-validator.js
```diff
@@ -40,6 +42,9 @@ function jsonSchemaValidatorFactory(
      * @return {Boolean} validation pass or throw error
      */
     JsonSchemaValidator.prototype.validate = function (schema, data) {
+
+        sanitizer.decrypt(data);
```
yyscamper: The `decrypt` should not be put in json schema validator. The validator is intented to validate whether the data conforms to the schema, it will not modify the data itself. So please find another reasonable place to put such logic.

DavidjohnBlodgett: @yyscamper Hey Felix, so I took your feedback and changed my other PR (on-tasks). During my testing, I had reverted on-core to master and was able to pass the JSON validation. So it would seem the bug resided in on-tasks, so I will close this PR (on-core)... since based on my testing, it seems to be not needed.

PR #222, opened 2016-10-21
Ssh obm service

srinia6:
Create inband-mgmt model. This PR includes:

```
Move sshSettings out of nodes.
Move the older obms model code to a base called as network-management.
create ibms collection.
```

@RackHD/corecommitters @brianparry @dalebremner @geoff-reid @BillyAbildgaard


rackhdmirror:

srinia6:
Jenkins:depends on RackHD/on-tasks#361

benbp:
👍

yyscamper:
Replied @brianparry 's comment in https://github.com/RackHD/on-core/pull/218 : 

My concern is that, if it is only because of the "obm" name itself that is not reasonable to include "ssh-ibm-service", it seems like an overkill to me to create a brand-new database model. Though the NetworkManagement helps the reduce the code duplication in on-core, but the data duplication in on-http is still too much.  The obm/ibm for me is just like we are trying to split the nodes model into compute-node-model, switch-node-model, pdu-node-model. So if the only concern is the "obm" name itself, why not copy the node model to add an attribute "type" in current "obm" model, using the type to distinguish obm and ibm (of course, the "obm" name itself need some refactor as well). If so, nearly all code can be shared.

For example:
```
{
    type: 'obm',
    service: 'ipmi-obm-service',
    config: {
        host: '172.31.128.9',
        user: 'admin',
        password: 'admin'
   }
}

{
    type: 'ibm',
    service: 'ssh-ibm-service',
    config: {
        host: '172.31.128.9',
        user: 'admin',
        password: 'admin'
   }
}
```

srinia6:
@yyscamper : Thanks Felix, makes sense. But we will have snmp under ibm collections too. The idea was the have snmp and ssh under ibm [which are NOT out-of-band ]collections. I think thats what @benbp eluded to in the previous PR, from his comments?  Seems like 2 different views/approach, unless i am understanding them differently. I will wait for input from @benbp and @brianparry

brianparry:
@yyscamper There is certainly some duplication in on-http, but I think this is mitigated pretty well by using services to do most of the work.  Most of the duplication is in the controllers, and yes, we could have reused more here but what would that really have achieved?  We would have eliminated a handful of near trivial functions but also introduced an exception to our pattern of mapping controllers to API endpoints.  We also would still have needed an ibm specific patch method.  IMO - this just makes things more confusing.  This is exactly the same pattern used for templates, profiles, and views.  Each of these is a renderable object with a common base model and its own collection.  Each has its own set of API routes and controllers.   There is some duplication here as well, but IMO there is little value in eliminating it.

lib/models/network-management.js
```diff
@@ -0,0 +1,229 @@
...
+            })
+            .then(function(mgmt) {
+                if (mgmt && reveal) {
+                    return _revealSecrets(mgmt);
+                }
+                return mgmt;
+            });
+        },
+
+        // This function 'upserts' a new OBM/IBM into the corresponding mgmt collection.
```
srinia6: ok

yyscamper:
@brianparry : replied in Slack channel.

yyscamper:
Sorry for duplicated comment, I don't know how to remove it.

srinia6:
test this please

PR #226, opened 2016-11-01
fix mongo indexes creation for all models

yyscamper:
This PR originates from ODR-892 (https://hwjiraprd01.corp.emc.com/browse/ODR-892), @cgx027 found two problems about mongo index:
(1) some index settings in our code doesn't take effect.
(2) lack index for some frequently operation

I also found the waterline official doc said there is some outstanding bug about index, see https://github.com/balderdashy/waterline-docs/blob/master/models/data-types-attributes.md#index, meanwhile, the waterline doesn't support compound index.

I see some models (such as obm/lookups) will manually create mongo index, but most other models don't do this way.

This PR is to let the mongo index creation take effect for all models. The index can be defined in following two ways:
(1) In waterline attributes, if the `index` is set to `true`, code will auto extract the index and create it regardless waterline's internal operation.
```
attributes: {
    name: {
        type: 'string',
        index: true
    }
}
```

(2) Use magic option `$indexes` for model, this will take precedent comparing with above. Meanwhile, this can be used to set compound index.
```
$indexes: [
    {
        keys: { foo: 1, bar: 1 }
        options: { unique: true }
    },
    {
        keys: { x: 1, y: 1, z: -1 }
    }
]
```

@RackHD/corecommitters @cgx027 @iceiilin @keedya @VulpesArtificem 

jenkins: depends on https://github.com/RackHD/on-taskgraph/pull/188

### DIFF
--- a/lib/models/lookup.js
+++ b/lib/models/lookup.js
@@ -17,19 +17,27 @@ LookupModelFactory.$inject = [
     '_'
 ];
 
-function LookupModelFactory (waterline, Model, assert, Errors, Promise, Constants, configuration, validator, _) {
+function LookupModelFactory (
+    waterline,
+    Model,
+    assert,
+    Errors,
+    Promise,
+    Constants,
+    configuration,
+    validator,
+    _
+) {
     var dbType = configuration.get('databaseType', 'mongo');
     return Model.extend({
         connection: dbType,
         identity: 'lookups',
         attributes: {
             node: {
-                model: 'nodes',
-                index: true
+                model: 'nodes'
             },
             ipAddress: {
                 type: 'string',
-                unique: true,
                 regex: Constants.Regex.IpAddress
             },
             macAddress: {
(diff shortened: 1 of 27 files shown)


iceiilin:
:+1:

yyscamper:
Hi @brianparry : Thanks for your comment. You are right, there may be conflict between waterline's index operation and our manual operation, so as you suggested, I have removed all waterline's index setting some attribute, and only allow our manually mongo index settings take effect.

cgx027:
:+1:

lib/models/lookup.js
```diff
@@ -58,7 +57,7 @@ function LookupModelFactory (
             },
             {
                 keys: { ipAddress: 1},
-                options: { unique: true }
+                options: { unique: false }
```
cgx027: @yyscamper , I would believe ipAddress should be unit in lookup table. When will it be NOT unique?

yyscamper: Thanks @cgx027 : I need to check @sunnyqianzhang 's system to see whether the duplicated ipAddress is sure added into RackHD, may be this is another problem. I will check before merging.

yyscamper: @cgx027 : I checked @sunnyqianzhang 's system, and found multiple lookup entries don't have ipAddress (ipAddress is not a required item for a lookup entry). So this is the reason why the ipAddress index need not be unique.

yyscamper:
@brianparry :  @sunnyqianzhang help to create the migration script: https://github.com/RackHD/on-tools/pull/46, I hope you can take a look at it.

sunnyqianzhang:
This change looks good to me +1

changev:
test this please

changev:
test this please

yyscamper:
jenkins: test this please

PR #228, opened 2016-11-14
Reverted Encryption at mongo store level, refactored Sanitizer Object.

DavidjohnBlodgett:
@RackHD/corecommitters 

Ok, so this guts the existing encryption I introduced... and I refactored the sanitizer object to serve up a single method that can be used to recursively redact secret fields... this PR will have a dependent PR within on-http to resolve the original issue:  https://github.com/RackHD/RackHD/issues/241

### DIFF
--- a/lib/common/model.js
+++ b/lib/common/model.js
@@ -14,8 +14,7 @@ ModelFactory.$inject = [
     'Errors',
     'Waterline',
     'Protocol.Waterline',
-    'Services.Waterline',
-    'Sanitizer'
+    'Services.Waterline'
 ];
 
 function ModelFactory (
@@ -27,8 +26,7 @@ function ModelFactory (
     Errors,
     Waterline,
     waterlineProtocol,
-    waterline,
-    sanitizer
+    waterline
 ) {
     var url = require('url');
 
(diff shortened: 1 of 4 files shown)


cgx027:
:100:

tldavies:
👍

lib/common/model.js
```diff
@@ -197,9 +195,6 @@ function ModelFactory (
             }
             var update = args.shift();
 
-            // encrypt password fields
-            sanitizer.encrypt(update);
```
benbp: I'm confused, is this encrypted elsewhere, or are we now just storing decrypted values in the database?

DavidjohnBlodgett: just in the database.

benbp: Ah ok

DavidjohnBlodgett:
@RackHD/corecommitters this should be merge ready.

DavidjohnBlodgett:
@brianparry so I was unable to uncover evidence to support my claim that encryption is happening before our secrets hit disk (aside from the code in https://github.com/RackHD/on-core/blob/d1444dc915c30fe3489caeb1cf8b32de37f98622/lib/models/obms.js#L42) 

That being said, since my first implementation of an encryption strategy has now blocked a number of people, and we have some more work to hash out a strategy we're all comfortable with, I'm recommending we merge this reversion and attempt to circle back on a new solution.

To directly answer your question, from what I can find... yes this change will save unecrypted password data to disk unless using the obm api, and yes the ultimate goal here is to just mask the secrets from the user in specific api calls.

PR #229, opened 2016-11-15
RAC-170: RackHD support for multiple AMQP servers

derrickostertag:
Modified RackHD on-core to allow an array of AMQP server IP addresses as an option during the AMQP connection call.

### DIFF
--- a/lib/common/messenger.js
+++ b/lib/common/messenger.js
@@ -92,8 +92,13 @@ function messengerFactory(
             options = {
                 url: uri
             };
-        } else {
+        } else if (Array.isArray(uri)) {
             options = uri;
+        } else {
+            console.log('Setting AMQP URL to localhost');
+            options = {
+                url: 'amqp://localhost'
+            };
         }
 
         var clientOptions = {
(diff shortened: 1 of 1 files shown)


derrickostertag:
@RackHD/corecommitters

lib/common/messenger.js
```diff
@@ -86,9 +86,16 @@ function messengerFactory(
...
-        var options = {
-            url: uri
-        };
+        var options;
+
+        if (typeof uri === 'string') {
+            options = {
+                url: uri
+            };
+        } else {
```
cgx027: Looks good. Just wondering if it would be a good idea to add add ```else if (Array.isArray(url))``` to explicitly declare that an Array is expected here, for code readability consideration. Looks good and :+1:  anyway.

derrickostertag:
I added a default case when not an array or a string.

derrickostertag:
@RackHD/corecommitters

larry-dean:
+1

derrickostertag:
@RackHD/corecommitters - This has two +1's

PR #231, opened 2016-12-09
External lookup helper support.

jeamland:
Adds the ability to call an external script to fill in missing lookup entries.

### DIFF
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -100,7 +100,7 @@ function lookupServiceFactory(
     };
 
     LookupService.prototype.resetMacRequests = function () {
-        if (configuration.get('externalLookupHelper', null) != null) {
+        if (configuration.get('externalLookupHelper', null)) {
             this.macRequests = {};
         } else {
             delete this.macRequests;
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -105,7 +105,7 @@ function lookupServiceFactory(
         } else {
             delete this.macRequests;
         }
-    }
+    };
 
     /**
      * Look up an entry in the lookups collection by IP. Optionally fall back
@@ -119,7 +119,7 @@ function lookupServiceFactory(
         .then(function() {
             return waterline.lookups.findOneByTerm(ip);
         });
-    }
+    };
 
     /**
      * Look up an entry in the lookups collection by MAC. Optionally fall back
@@ -142,7 +142,7 @@ function lookupServiceFactory(
             }
             throw error;
         }.bind(this));
-    }
+    };
 
     /**
      * Run an external helper script to fill in missing lookup entries.
@@ -166,7 +166,7 @@ function lookupServiceFactory(
         return this.processHelperResults(helper).then(function () {
             this.handleHelperPromises(mac)
         }.bind(this));
-    }
+    };
 
     /**
      * Process the output from an external lookup helper script.
@@ -190,9 +190,9 @@ function lookupServiceFactory(
                 promises.push(this.setIpAddress(items[1], items[0]));
             }.bind(this))
 
-            return Promise.all(promises)
+            return Promise.all(promises);
         }.bind(this));
-    }
+    };
 
     /**
      * Clean up remaining helper script promises if any are waiting on the
@@ -206,7 +206,7 @@ function lookupServiceFactory(
         requests.forEach(function (responder) {
             responder.resolve();
         });
-    }
+    };
 
     /**
      * Reset the cache for a specific IP or MAC address.
--- a/lib/services/lookup.js
+++ b/lib/services/lookup.js
@@ -13,7 +13,8 @@ lookupServiceFactory.$inject = [
     'Errors',
     '_',
     'lru-cache',
-    'ARPCache'
+    'ARPCache',
+    'ChildProcess'
 ];
 
 /**
@@ -36,7 +37,8 @@ function lookupServiceFactory(
     Errors,
     _,
     LRU,
-    arpCache
+    arpCache,
+    ChildProcess
 ) {
     // TODO: stolen from http-service could stand to be a re-usable helper.
     /**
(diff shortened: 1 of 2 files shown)


brianparry:
ok to test

brianparry:
add to whitelist

lib/services/lookup.js
```diff
@@ -96,6 +99,115 @@ function lookupServiceFactory(
...
+     * Each entry will be turned into arguments for this.setIpAddress.
+     * @private
+     * @param {helper} The ChildProcess object representing the helper script.
+     */
+    LookupService.prototype.processHelperResults = function (helper) {
+        var self = this;
+        return helper.run().then(function (ret) {
+            var promises = [];
+
+            ret.stdout.split('\n').forEach(function (line) {
```
brianparry: @jeamland Just as an FYI - you can use Promise.map in place of a promise array / Promise.all.  http://bluebirdjs.com/docs/api/promise.map.html

brianparry:
Since the Jenkins failure appears to be in an area unrelated to this change: jenkins, test this please

tannoa2:
jenkins, test this please

tannoa2:
jenkins, test this please

tannoa2:
test this please

anhou:
@jeamland no big concern except security @yyscamper mentioned,  at present, there's no customized script called by RackHD except some specific tools such as ipmitool/snmp tools.
@yyscamper this PR's purpose has no relationship with the issue @lanchongyizu resolve. @jeamland could provide more background about this.

jeamland:
So in our use case RackHD is not running on our DHCP server. Our DHCP server is a series of Infoblox appliances deployed in our labs. The situation I've seen multiple times while developing our RackHD-based setup is that RackHD can't do a lookup to find a BMC IP address based on its MAC. This happens because the BMCs aren't getting their IP addresses in a way that the RackHD server can see. It doesn't have the lease data from the DHCP server and it additionally doesn't have local ARP entries to parse. The end result being that any attempt to do, for example, an IPMI-based reboot fails as RackHD only has a MAC address, can't look up the corresponding IP via any mechanism it has available to it and thus the operation fails.

The external helper solution to this allows me to plug in Infoblox with a mechanism that could also be used by others fairly trivially. I understand your security concern but as @anhou points out it's no worse than your existing exposure to ipmitool. In fact I'm using a ChildProcess helper object that's part of your codebase.

I'm also happy if you want to develop a plugin solution that I could write something for but it would have to work in such a way that I could install RackHD from vanilla source (I'm currently using the .deb packages) and then point RackHD at my plugin using configuration variables. This solution allows me to do just that.

keedya:
test this please

keedya:
test this please

keedya:
test this please

keedya:
test this please

keedya:
test this please

tannoa2:
test this please

brianparry:
@keedya Took down the problematic build slave until we can figure out what's wrong with it.  Next test run should be on slave 02 which is working.  Jenkins: test this please

tannoa2:
test this please

tannoa2:
test this please

yyscamper:
@jeamland : We had internally discussed the plugin mechanism to let user to inject some special node, but even the proposal has not been public yet, so this is still a question whether we will eventually choose this way and even we agree, there is still a long way to let it done.

In your case, whether the BMC IP is dynamically changed? If not, that means if a BMC is always assigned a dedicated IP, how about the pre-load all IP/MAC relation into RackHD via `POST /lookups` API?

jeamland:
The BMC IPs are currently dynamic and there aren't any plans to make them static.

yyscamper:
@jeamland : Is there any possibility to inject an action in Infoblox to call RackHD API to inject the new lookup if there is any IP assigned or changing?

What I am trying to avoid is to call an external script directly. There is another way to inject external behavior is via callback-url. For example, set a lookup callback-url in config.json as "http://foo/mylookup". If  RackHD internal lookup fails, then fall back to send request to callback-url via same query, the callback-url can return something back to RackHD. 

We have another feature that enable user to input a callback url to receive the RackHD event notification, so I propose the same design to this lookup problem.

Of course, comparing callback-url with your externel script. The callback-url is a little complicated in user side, because user need to setup a HTTP server to listen the callback request. The benefit is to resolve the security concern and it also has clearer and better interface design.

How do you think? Add @brianparry and @anhou for comment.

jeamland:
No, Infoblox doesn't have anything like that. I checked. Also Infoblox is handling a lot more than the systems RackHD would be looking after so RackHD would be getting a huge amount of data it didn't need and that's before we take into account that there will be multiple RackHD instances managing specific groups of systems.

I get where you're coming from on the security angle but this is a script that the user installs and that the user configured in to RackHD as an optional thing. It's no more a security hazard than calling out to ipmitool which RackHD already does. The Infoblox script is incredibly simple and is something I don't have to monitor or manage.

anhou:
@yyscamper In general, @jeamland 's opinions make sense to me, 'callback url' is not enough to achieve this function. and the script has the same security level compared with ipmitool after deployment.  
@jeamland one thing is that could you also contribute the script. I think infobox as one of the DHCP service, the script to get mac/ip also could be referenced by other DHCP services or used by others. otherwise, it looks like it has a 'hole' (has interface but no content) in RackHD. infobox could be the first to do this and guide others.

Actually, except security, RackHD call python or any other language scripts is similar with calling local solid interface which accept mac/ip information. the difference is 1. the language could be different 2. the interface accept inputs from script result, not internal codes or functions. I have no strong opinion to achieve this function inside RackHD.  so this PR is 👍  for me

brianparry:
@yyscamper You make a fair point about the security risks of shelling out to an external script. Maybe a naive opinion, but I think we need to have a degree of trust that users with sufficient privilege to install and configure rackhd aren't going to do malicious things.  And, as @jeamland points out, this is no less secure than other things we do already, so :+1: from me.

yyscamper:
Thanks for all your explanation, I think you are right, we should trust user will do right thing. So I will withdraw my disapproval.

PR #236, opened 2017-01-05
The first version of splitting external events with different publish method

sunnyqianzhang:
This is the first version, what it does:
1.  Wrap the messenger.js in lib/common
2.  Add two different methods for publishing the internal and external events 

What needs to be donw:
1. Add comments to the methods
2. Removing the publish mehtod and replace it with  publish internal and external methods respectively.
3. Redefine publish external events to complete the discussed format
 @iceiilin @pengz1 @anhou @yyscamper  
 
The second version is done:
what it does:
1.  Remove the publish method on lib/service/messenger.js
2. Change the publish method to publishExternal/publishInternal in all repo. 
Code change as item 2 is done on other repos too, but doesn't send out for review now

### DIFF
--- a/lib/services/messenger.js
+++ b/lib/services/messenger.js
@@ -6,9 +6,87 @@ module.exports = messengerServiceFactory;
 
 messengerServiceFactory.$provide = 'Services.Messenger';
 messengerServiceFactory.$inject = [
-    'Messenger'
+    'Messenger',
+    'Assert'
+
 ];
 
-function messengerServiceFactory(Messenger) {
-    return new Messenger();
+function messengerServiceFactory(
+    Messenger,
+    assert
+) {
+    function MessengerService(){
+        this.messenger = new Messenger();
+    }
+
+    MessengerService.prototype.validate = function validate(data) {
+	return this.messenger.validate(data);
+    };
+
+    MessengerService.prototype.queueName = function queueName(name, routingKey){
+	return this.messenger.queueName(name, routingKey);
+    };
+
+    MessengerService.prototype.start = function start() {
+        return this.messenger.start();
+    };
+
+    MessengerService.prototype.stop = function stop() {
+        return this.messenger.stop();
+    };
+
+    MessengerService.prototype.publishExternalEvents  = function(
+        name, routingKey, data, options
+        ) {
+            assert.ok(data.hasOwnProperty("type"));
+            assert.ok(data.hasOwnProperty("action"));
+            assert.ok(data.hasOwnProperty("severity"));
+            assert.ok(data.hasOwnProperty("typeId"));
+            assert.ok(data.hasOwnProperty("nodeId"));
+            assert.ok(data.hasOwnProperty("payload"));
+
+            assert.string(data.type);
+            assert.string(data.action);
+            assert.string(data.serverity);
+            assert.string(data.typeId);
+            if (data.nodeId !== null){
+                assert.string(data.nodeId);}
+            assert.object(data.payload);
+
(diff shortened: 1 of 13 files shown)


lib/services/messenger.js
```diff
@@ -6,9 +6,67 @@ module.exports = messengerServiceFactory;
...
+		return this.messenger.safeSubscriptionDispose(context);
+	}
+
+   MessengerService.prototype.subscribeCallback = function(context, options) {
+	   return this.messenger.subscribeCallback(context, options);
+   }
+    
+    return new MessengerService();
 }
+
```
pengz1: Actually codes here are redundant, you just need to add publishInternalEvents/publishExternalEvents which call this.publish is enough. And the most important things are:
Where to add hooks?
Where to align common events format?

Those two items are the important items to differentiate internal and external events.

Minor question: change publishInternalEvents/publishExternalEvents to publishExt/publishInt will be better to be used.

lib/workflow/stores/mongo.js
```diff
@@ -343,6 +343,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
         var query = {
             instanceId: data.graphId
         };
+        //console.log("\n\n\n&&&&&&&&&&&&&&&&&instanceId&&&&&&&&&&&&&&&&\n\n\n",data.graphId)
```
iceiilin: this file can be removed from git commit

lib/services/messenger.js
```diff
@@ -6,9 +6,67 @@ module.exports = messengerServiceFactory;
...
+    };
+
+
+	MessengerService.prototype.safeSubscriptionDispose = function (context) {
+		return this.messenger.safeSubscriptionDispose(context);
+	}
+
+   MessengerService.prototype.subscribeCallback = function(context, options) {
+	   return this.messenger.subscribeCallback(context, options);
+   }
```
iceiilin: this function is for internal use, and doesn't need to expose

lib/services/messenger.js
```diff
@@ -6,9 +6,67 @@ module.exports = messengerServiceFactory;
...
+    };
+
+    MessengerService.prototype.request = function(name, routingKey, data, type, timeout) {
+        return this.messenger.request(name, routingKey, data, type, timeout);
+    };
+
+
+	MessengerService.prototype.safeSubscriptionDispose = function (context) {
+		return this.messenger.safeSubscriptionDispose(context);
+	}
```
iceiilin: this function is for internal use, and doesn't need to expose

lib/services/messenger.js
iceiilin: using inherits is another solution that needs consideration, which could save some efforts on expose messenger interfaces. If designed in this way, common/messenger.js is the "first priority" in this file, while webhook is instantiated as a variable in this class, which might be different from the suggestion of @yyscamper 

hooks could be added in publishExternalEvents(), and common events format is marshaled in this function as well.

sunnyqianzhang: I thought of inherits at first, but according to @yyscamper the relationship between this messenger.js and the messenger.js in lib/common are not in inherits relationships, being an instance of it is a better choice. As for the webhook, it could be treated in a similar way: in publishExternalEvents, this.hook = new Hook() or something like that.

sunnyqianzhang: There might be some method defined in lib/common/messenger.js, which is not used and thus not necessary to redefine here, yet I plan to leave it as it is now, and remove the unused one when another task at hand is finished.

PR #238, opened 2017-01-09
[WIP] Hook-code

pengz1:
Create hook model and hook services for hook feature.
Hook services and AMQP events are supposed to be combined and this work is still ongoing, please ignore service related function in this PR.

### DIFF
--- a/lib/models/hooks.js
+++ b/lib/models/hooks.js
@@ -1,0 +1,43 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = HooksModelFactory;
+
+HooksModelFactory.$provide = 'Models.Hooks';
+HooksModelFactory.$inject = [
+    'Model',
+    'Services.Configuration',
+    'uuid'
+];
+
+function HooksModelFactory (Model, configuration, uuid) {
+
+    return Model.extend({
+        connection: configuration.get('databaseType', 'mongo'),
+        identity: 'hooks',
+        attributes: {
+            id: {
+                type: 'string',
+                uuidv4: true,
+                primaryKey: true,
+                unique: true,
+                required: true,
+                defaultsTo: function() { return uuid.v4(); }
+            },
+            name: {
+                type: 'string',
+                required: false
+            },
+            url: {
+                type: 'string',
+                required: true
+            },
+            filters: {
+                type: 'array',
+                required: false,
+                defaultsTo: []
+            }
+        }
+    });
+}
--- a/lib/services/hook.js
+++ b/lib/services/hook.js
@@ -1,0 +1,104 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+var http =  require('http');
+var https =  require('https');
+var url = require('url');
+
+module.exports = hookServiceFactory;
+
+hookServiceFactory.$provide = 'Services.Hook';
+hookServiceFactory.$inject = [
+    'Services.Messenger',
+    '_',
+    'Assert',
+    'Services.Waterline'
+];
+
+function hookServiceFactory(
+    messenger,
+    _,
+    assert,
+    waterline
+) {
+    function HookService(){
+    }
+    
+    /**
+     * Publishes the JSON serialized from 'data' into AMQP based on exchange and routingKey.
+     *
+     * @param {string} name
+     * @param {string} routingKey
+     * @param {Object} data
+     * @param {Object} options
+     * @returns {Promise}
+     */
+    HookService.prototype.publish = function(name, routingKey, data, options) {
+        assert.string(name, 'name');
+        assert.string(routingKey, 'routingKey');
+        assert.object(data, 'data');
+        return messenger.publish(name, routingKey, data, options)
+        .tap(function(){
+            return waterline.hooks.find({})
+            .then(function(hookList){
+                data.routingKey = routingKey;
+                postDataToUrls(hookList, data);
+            });
+        });
+    };
+
+    function postDataToUrls(hookList, data) {
+        if(_.isEmpty(hookList)) {
+            return;
+        }
+        _.forEach(hookList, function(hook){
+            if(_validateUrlRule(data)){
+                _postDataToUrl(hook, data);
+            }
+        });
+    }
+
+    function _validateUrlRule(data) {
+        //TODO: design rules here.
+        return true;
+    }
+
+    function _postDataToUrl(hook, data){
+        assert.object(hook);
+        assert.object(data);
+        assert.string(hook.url);
+        
+        var hookObj = url.parse(hook.url);
+        var options = {
+            hostname: hookObj.hostname,
+            port: hookObj.port,
+            path: hookObj.path,
+            method: 'POST',
+            headers: {'Content-Type': hook.contentType || 'application/json'}
+        };
+        var req;
+        if ( hookObj.protocol === "https:") {
+            req = https.request(options, function(res){
+                res.setEncoding('utf8');
+            });
+        } else {
+            req = http.request(options, function(res){
+                res.setEncoding('utf8');
+            });
+        }
+        req.write(JSON.stringify(data));
+        req.end();
+    }
+
+    HookService.prototype.start = function start() {
+        this.started = true;
+        return Promise.resolve();
+    };
+
+    HookService.prototype.stop = function stop() {
+        return Promise.resolve();
+    };
+
+    return new HookService();
+}
--- a/lib/protocol/events.js
+++ b/lib/protocol/events.js
@@ -9,6 +9,7 @@ eventsProtocolFactory.$inject = [
     'Constants',
     '_',
     'Services.Messenger',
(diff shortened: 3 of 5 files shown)


pengz1:
@yyscamper @lanchongyizu @iceiilin @anhou

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+        req.end();
+    }
+
+    HookService.prototype.start = function start() {
+        this.started = true;
+        return Promise.resolve();
+    };
+
+    HookService.prototype.stop = function stop() {
+        return Promise.resolve();
```
brianparry: Set `this.started = false`?

lib/models/hooks.js
```diff
@@ -0,0 +1,43 @@
...
+            name: {
+                type: 'string',
+                required: false
+            },
+            url: {
+                type: 'string',
+                required: true
+            },
+            filters: {
+                type: 'array',
```
brianparry: Is the intent here to allow a list of rules that, when satisfied, would trigger the hook?

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+     * Publishes the JSON serialized from 'data' into AMQP based on exchange and routingKey.
+     *
+     * @param {string} name
+     * @param {string} routingKey
+     * @param {Object} data
+     * @param {Object} options
+     * @returns {Promise}
+     */
+    HookService.prototype.publish = function(name, routingKey, data, options) {
+        assert.string(name, 'name');
```
brianparry: These asserts should be enclosed within a `Promise.try` block to avoid throwing a synchronous error.

lib/models/hooks.js
```diff
@@ -0,0 +1,43 @@
...
+            id: {
+                type: 'string',
+                uuidv4: true,
+                primaryKey: true,
+                unique: true,
+                required: true,
+                defaultsTo: function() { return uuid.v4(); }
+            },
+            name: {
+                type: 'string',
```
brianparry: Are multiple hooks with the same name allowed?

lib/models/hooks.js
yyscamper: Yes, `filters` is an array of a list of rules, if any rule is satisfied, then the event will be pushed into the hook via the hook url.

This is something like user may have interest on multiple topics (node, poller and graph), set the `filters` as `array` will give user the ability to subscribe multiple topics.

lib/models/hooks.js
pengz1: @brianparry  Hook is supposed to be identified by url or id. Name is just an optional parameter set by customer and we will not identify name itself.

yyscamper: Yes, the `name` is only some words used by user to distinguish each hook. Since each hook is bind with a url, but the url sometimes is not so distinguish, such as `http://192.168.290.110:8099/test/hooks` to subscribe node topic, and use `http://192.168.290.123:8119/hooks` to subscribe poller topic, these two urls is not so easy to distinguish.


How to name a hook is owned by user, user can call it whatever he/she wants, so even a duplicated name is allowed and even user can choose not input any `name`.

lib/models/hooks.js
brianparry: @yyscamper Got it.  That makes sense.

lib/models/hooks.js
pengz1: @brianparry 
Tag is already used by RackHD, I am a little concerned to used this word on hook's attribute. Actually our current node/sku also have an attribute "name" that could be used by user. I suggest we can keep this attribute "name" unless we find a better naming word.

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+        assert.object(data);
+        assert.string(hook.url);
+        
+        var hookObj = url.parse(hook.url);
+        var options = {
+            hostname: hookObj.hostname,
+            port: hookObj.port,
+            path: hookObj.path,
+            method: 'POST',
+            headers: {'Content-Type': hook.contentType || 'application/json'}
```
yyscamper: I admit supporting multiple contentType will be a good feature. But that means you have to write some code to change the JSON data to different content type. But obviously you don't write such code.

My suggestion is support "application/json" in your first PR, then you can write an additional PR to support other contentType.

Meanwhile, don't forget to add the `contentType` into the hook model if you decide to implement this feature.

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+        //TODO: design rules here.
+        return true;
+    }
+
+    function _postDataToUrl(hook, data){
+        assert.object(hook);
+        assert.object(data);
+        assert.string(hook.url);
+        
+        var hookObj = url.parse(hook.url);
```
yyscamper: This variable name is not so accurate, this is not a hook object, just an object for hook url

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+            method: 'POST',
+            headers: {'Content-Type': hook.contentType || 'application/json'}
+        };
+        var req;
+        if ( hookObj.protocol === "https:") {
+            req = https.request(options, function(res){
+                res.setEncoding('utf8');
+            });
+        } else {
+            req = http.request(options, function(res){
```
yyscamper: Actually you don't handle the `end` and `error` events, this will be not easy to debug.

For example, if the http request fails, you'd better print some warning log, so that it assists debugging.
If the http request succeeds, but the server responds 500 other error, that means the event is still not successfully delivered to user. Some warning log will still assist debugging.

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+            }
+        });
+    }
+
+    function _validateUrlRule(data) {
+        //TODO: design rules here.
+        return true;
+    }
+
+    function _postDataToUrl(hook, data){
```
yyscamper: This while function is not wrapped as Promise, that means any uncaught error will directly lead the whole service crashes.

To catch or report error, I suggesting using promise.

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+     */
+    HookService.prototype.publish = function(name, routingKey, data, options) {
+        assert.string(name, 'name');
+        assert.string(routingKey, 'routingKey');
+        assert.object(data, 'data');
+        return messenger.publish(name, routingKey, data, options)
+        .tap(function(){
+            return waterline.hooks.find({})
+            .then(function(hookList){
+                data.routingKey = routingKey;
```
yyscamper: Why you add the routingKey in data, routingKey is only specific for AMQP channel. All event format is defined in @anhou 's common event format, you should avoid adding some additional fields.

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Publishes the JSON serialized from 'data' into AMQP based on exchange and routingKey.
+     *
+     * @param {string} name
+     * @param {string} routingKey
+     * @param {Object} data
+     * @param {Object} options
+     * @returns {Promise}
+     */
+    HookService.prototype.publish = function(name, routingKey, data, options) {
```
yyscamper: The AMQP messenger logic is put within the hook, but actually they are in equal status, so you should remove all AMQP messenger code out of the hookService.

@sunnyqianzhang has a PR to split internal and external events, bases on her PR, I think the better place to combine the hook and AMQP messenger is "Services.Messenger"

lib/services/hook.js
```diff
@@ -0,0 +1,104 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
```
yyscamper: Overall, hook is better not to be a service, in my thinking, the difference between Hook, Messenger and Services.Messenger is as following:
(1) For the term "Messenger", it is a kind of channel that supports publish/subscribe message. The channel can be AMQP or Hook.
(2) The current di module `Messenger` is better to rename to "Messenger.AMQP", it is a class that publish/subscribe message via AMQP channel.
(3) The Hook can be casted to a messenger, it is a class that publish/subscribe message via Web URL. But in my vision of Hook, it can do more than a messenger. So Hook is not equal to a messenger but it can behave as a messenger.
(4) The "Services.Messenger" is to manage all message channels, that means all AMQP messenger and Hook Messenger instances should be initialized and registered in Services.Messenger.

Add @sunnyqianzhang @iceiilin @anhou

lib/models/hooks.js
brianparry: @pengz1 My criticism here was mainly around the upsert-like create method you had in your on-http PR which expected names to be unique.  Since you've changed this, I think the proposed semantics and implementation should be fine and are inline with the conventions we've used elsewhere.

lib/models/hooks.js
```diff
@@ -0,0 +1,43 @@
...
+                required: false
+            },
+            url: {
+                type: 'string',
```
brianparry: @pengz1 This should be unique since the create method in your on-http PR uses this to detect duplicate hooks.

PR #241, opened 2017-01-12
fix houndci parse error by removing comment in jshintrc

yyscamper:
The houndci doesn't strip  the comment from the jshintrc file (a JSON file), so it complains an error about "fail to parsing config file".

It is true that the standard JSON doesn't support comment, so my fix is to completely remove the comment from jshintrc file.

@anhou @iceiilin @pengz1

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,7 +1,6 @@
 {
-    "maxerr"        : 50,     //Max errors before it fails the build
+    "maxerr"        : 50,
 
-    //Enforcing
     "bitwise" : true,
     "camelcase" : true,
     "curly" : true,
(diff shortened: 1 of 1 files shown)


pengz1:
+1

iceiilin:
+1

anhou:
@yyscamper  why it could parse .jshintrc successfully previously?

yyscamper:
@anhou : I had reported the same bug for houndci (see https://github.com/houndci/hound/issues/1040 ), they made a fix to remove the comment before parse the .jshintrc JSON file.

But I notice recently houndci has code architecture change, they should make a mistake for this problem again.

Comparing to ask houndci  team to fix this, I'd rather we clean our config file format, since the comment is surely not supported in standard JSON.

anhou:
@yyscamper Thanks for clarification, it's 👍 from me. and  other repos also has the same problem. could you also submit the same PRs to other repos, let's fix them quickly today.

PR #243, opened 2017-01-13
nonterminal, handled tasks no longer prevent graph from finishing

VulpesArtificem:
solves https://rackhd.atlassian.net/browse/RAC-1938

changed the mongo store checkGraphSucceeded method to succeed on graphs where the only task dependencies we find with waterline.taskdependencies.find are non-terminal.
@rolandpoulter @pscharla


tldavies:
👍

PR #244, opened 2017-01-13
 Implement hook feature

pengz1:
Jenkins depends on: https://github.com/RackHD/on-tasks/pull/383
 * Hook features need database model and method to post data to user
   provided url
 * Thus hook model is created as well as hook publish method
 * However hook publish method leverage http tool, http tool is moved from
   on-task repo to lib/common

### DIFF
--- a/lib/common/hook.js
+++ b/lib/common/hook.js
@@ -65,5 +65,12 @@ function hookFactory(
         return true;
     }
 
-    return new Hook();
+    Hook.prototype.start = function start() {
+        return Promise.resolve();
+    };
+
+    Hook.prototype.stop = function stop() {
+        return Promise.resolve();
+    };
+    return Hook;
 }
--- a/lib/common/hook.js
+++ b/lib/common/hook.js
@@ -9,14 +9,16 @@ hookFactory.$inject = [
     '_',
     'Assert',
     'Services.Waterline',
-    'HttpTool'
+    'HttpTool',
+    'Promise'
 ];
 
 function hookFactory(
     _,
     assert,
     waterline,
-    HttpTool
+    HttpTool,
+    Promise
 ) {
     function Hook(){
     }
@@ -29,7 +31,9 @@ function hookFactory(
      */
     Hook.prototype.publish = function(data) {
         var self = this;
-        return waterline.hooks.find({})
+        return Promise.try(function(){
+            return waterline.hooks.find({});
+        })
         .map(function(hook){
             return self._publish(hook, data);
         });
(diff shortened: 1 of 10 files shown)


pengz1:
@yyscamper @brianparry @sunnyqianzhang @anhou @iceiilin 
As there are too much changes on WIP code, I created a new PR. Here is the old PR and I am going to close it: https://github.com/RackHD/on-core/pull/238

lib/common/hook.js
```diff
@@ -0,0 +1,69 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = hookFactory;
+
+hookFactory.$provide = 'Hook';
```
yyscamper: Please rename to "Hooks" as this code is to handle all hooks rather than a particular hook.

spec/lib/common/hook-spec.js
```diff
@@ -0,0 +1,91 @@
...
+                    })
+                    .post('/test')
+                    .reply(201, 'OK');
+        findStub.withArgs({}).resolves([hooks[0]]);
+        return hook.publish(payload)
+        .then(function(){
+            expect(hook._publish).to.have.been.calledOnce;
+            expect(hook._publish).to.have.been.calledWith(hooks[0], payload);
+            expect(waterline.hooks.find).to.have.been.calledOnce;
+            nock.removeInterceptor(scope);
```
yyscamper: This should be called whatever the publish fail or success, so you can move it to [`finally`](http://bluebirdjs.com/docs/api/finally.html)

lib/common/hook.js
pengz1: I'd like to keep this hook, which is similar to messenger without 's'.

pengz1:
test this please

iceiilin:
:+1:

spec/lib/common/hook-spec.js
```diff
@@ -0,0 +1,117 @@
...
+        _publishSpy.reset();
+        findStub.reset();
+        nock.cleanAll();
+    });
+
+    it('should post data to hook url with http service', function () {
+        nock('http://172.1.1.1:8080')
+            .filteringRequestBody(function(body){
+                if (!_.isEqual(payload, JSON.parse(body))) {
+                    throw new Error('body is not matched');
```
iceiilin: is it better to use done(error) here? so the whole test won't be terminated, but only this case will be marked as fail.

panpan0000:
Please rebase your PR branch to latest master , otherwise, on-tasks unit-test will fail. (task-schema)

changev:
test this please

changev:
test this please

PR #245, opened 2017-01-17
Refactor publish external events to add several fields

sunnyqianzhang:
1. On lib.protocol, rename publishevent to publishExternalEvent
Check required field for external events in this function
2. Make changes to node event and heartbeat event, so that it conforms to the required fields for external event
@pengz1 @iceiilin @anhou @yyscamper

### DIFF
--- a/lib/protocol/events.js
+++ b/lib/protocol/events.js
@@ -11,8 +11,7 @@ eventsProtocolFactory.$inject = [
     'Services.Messenger',
     'Promise',
     'validator',
-    'Assert',
-    'Errors'
+    'Assert'
 ];
 
 function eventsProtocolFactory (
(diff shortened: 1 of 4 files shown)


lib/protocol/events.js
```diff
@@ -275,18 +273,29 @@ function eventsProtocolFactory (
...
+        assert.string(data.typeId);
+        assert.ok(data.hasOwnProperty("payload"));
+        if (data.nodeId !== null){
+            assert.string(data.nodeId);}
+
+        data.version = '1.0';
+        data.createdAt = new Date();
+
+        var routingKey = data.type + '.' + data.typeId + '.' + data.action +
+                         '.' + data.severity + '.' + data.nodeId;
```
pengz1: In this case, if nodeId is null or undefined, routingKey will be somthing like a.b.c.d.null or a.b.c.d.undefined, is this expected?

anhou: @pengz1 I'll fix this problem in another PR

PR #249, opened 2017-01-22
Implement hook filter feature

pengz1:
* To let user subscribe interested data, filter is designed based on event message header for hook feature.
     * User filter should be a Javascript style RE expression as a string in the design.
     * All data will be filtered against user defined hook filters and only qualified data will be posted to webhook url.
     * Each hook may have an array of filters, any filter condition can make an event eligible and sent to the hook.

### DIFF
--- a/lib/common/hook.js
+++ b/lib/common/hook.js
@@ -47,7 +47,10 @@ function hookFactory(
      * @returns {Promise}
      */
     Hook.prototype._publish = function(hook, data) {
-        if(!_isQualifiedData(data, hook.filters)){
+        //Current filtering mechanism only filters against common event header
+        //Payload data filtering in event messages is not supported
+        var isQualified = _filter(data, hook.filters);
+        if(!isQualified){
             return;
         }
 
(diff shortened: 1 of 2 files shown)


pengz1:
@lanchongyizu @yyscamper @anhou @linlynn @sunnyqianzhang

panpan0000:
Please rebase your PR branch to latest master , otherwise, on-tasks unit-test will fail. (task-schema)

lib/common/hook.js
```diff
@@ -0,0 +1,89 @@
...
+            isQualified = _conditionFilter(data, filter);
+            return !isQualified;
+        });
+        return isQualified;
+    }
+
+    function _conditionFilter(data, filter) {
+        var isPassed = true;
+        _.forEach(filter, function(value, key){
+            var pattern = new RegExp(value);
```
yyscamper: We add a TODO item here for performance concern, and hopefully it can be addressed in another PR:

Since after the hook is defined, the filter criteria isn't changed frequently (comparing with the occurrence frequency of event). So if the regular expression can be compiled in advance, then the filter directly uses the complied one, then the regular expression testing will save a lot time.

pengz1: @yyscamper  new RegExp() expression will create a regular expression of complied.

sunnyqianzhang: This PR is inconsistent with that in:
https://github.com/RackHD/on-core/pull/244

In this PR, no lib/service/hook.js is created, while in above PR it is. Which one will be finally checked in?

lib/common/hook.js
```diff
@@ -64,9 +67,28 @@ function hookFactory(
         });
     };
 
-    function _isQualifiedData(data, filters) {
-        //TODO: design filters for hook here.
-        return true;
+    function _filter(data, filters) {
```
iceiilin: suggest to add some comments for these two new functions for reader's better understanding, about the filter's structure, or why it is an OR logic while the other is an AND logic.

pengz1: How about user document?

iceiilin: thanks for you info. I am ok with that.

PR #251, opened 2017-01-24
Change check for multiple amqp URIs from an array to an object.

derrickostertag:
Multiple AMPQ servers are passed to amqp library as an array inside an object.
This changes the check of the uri parameter as an object instead of an array.


derrickostertag:
@johren @patelb10

johren:
test this please

PR #252, opened 2017-01-25
adding ibms collection

BillyAbildgaard:
merging from feature/ibms into master

@RackHD/veyron


lib/models/node.js
```diff
@@ -73,10 +73,9 @@ function NodeModelFactory (
             json: true,
             required: false
         },
-        sshSettings: {
```
dalebremner: Should the "snmpSettings" property be removed too?

PR #258, opened 2017-02-17
To add a function to get all macs and ip addresses from nodeId

sunnyqianzhang:
@pengz1 @iceiilin


lib/services/lookup.js
```diff
@@ -482,6 +482,31 @@ function lookupServiceFactory(
     };
 
     /**
+     * Returns a list of object containing IP and Mac Addresses for the given Node ID.
+     * @param  {String} id Node ID
+     * @return {Promise.<Array>} A promise fulfilled with an array of
+     * objects containing IP Addresses and Mac Addresses associated with the Node.
+     */
+    LookupService.prototype.findIpMacAddresses = function (id) {
```
yyscamper: A little suggestion, change the variable name from `id` to `nodeId` will be better, since the the caller may confuse the id is the lookup document id.

lib/services/lookup.js
```diff
@@ -482,6 +482,31 @@ function lookupServiceFactory(
...
+        var self = this;
+        assert.isMongoId(id);
```
yyscamper: Please remove this assertion, as RackHD will support multiple database, the nodeId will be changed to a common format  (such as uuid) rather than the ID that auto-generated by mongodb.

lib/services/lookup.js
```diff
@@ -482,6 +482,31 @@ function lookupServiceFactory(
...
+
+        return self.validateArpCache()
+        .then(function() {
+            return waterline.lookups.findByTerm(id).then(function (records) {
+                return _.reduce(records, function (ipMacAddresses, record) {
+		    var ipMacAddress = {};
+	            ipMacAddress.ipAddress = record.ipAddress;
+		    ipMacAddress.macAddress = record.macAddress;
```
yyscamper: indent problem

lib/services/lookup.js
```diff
@@ -482,6 +482,31 @@ function lookupServiceFactory(
...
+     * @return {Promise.<Array>} A promise fulfilled with an array of
+     * objects containing IP Addresses and Mac Addresses associated with the Node.
+     */
+    LookupService.prototype.findIpMacAddresses = function (id) {
+        var self = this;
+        assert.isMongoId(id);
+
+        return self.validateArpCache()
+        .then(function() {
+            return waterline.lookups.findByTerm(id).then(function (records) {
```
yyscamper: I don't think it's a good idea to use `findByTerm` as the `id` here is absolutely node Id. So use `waterline.lookups.find({ node: id})` will not only gain performance but also avoid a potentially bug.
If you look at the implementation of `findByTerm`, there is some 'smart` logic inside it, the input `id` may be an IP address, a MAC address or a nodeId, this function smartly detects the format of your input. But in your case, this kind of 'smart detection' is not necessary and may cause the bug that it thinks the id is MAC address rather than node id (especially you set a random node id in unit-test).

lib/services/lookup.js
```diff
@@ -482,6 +482,31 @@ function lookupServiceFactory(
...
+                return _.reduce(records, function (ipMacAddresses, record) {
+		    var ipMacAddress = {};
+	            ipMacAddress.ipAddress = record.ipAddress;
+		    ipMacAddress.macAddress = record.macAddress;
+                    ipMacAddresses.push(ipMacAddress);
```
yyscamper: Just a little coding suggestion, you could avoid to explicitly declare a varaible `ipMacAddress`, just do something like below:
```
ipMacAddresses.push({
    ipAddress: record.ipAddress,
    macAddress: record.macAddress
});
```

or event a more simple one:
```
ipMacAddresses.push(_.pick(record, ['ipAddress', 'macAddress']));
```

spec/lib/services/lookup-spec.js
```diff
@@ -597,19 +597,40 @@ describe('Lookup Service', function () {
...
-            expect(WaterlineService.lookups.setIp).to.have.been.calledOnce;
-            expect(WaterlineService.lookups.setIp).to.have.been.calledWith('ip', 'mac');
-        });
+            .then(function() {
+                expect(WaterlineService.lookups.setIp).to.have.been.calledOnce;
+                expect(WaterlineService.lookups.setIp).to.have.been.calledWith('ip', 'mac');
+            });
     });
+
+    it('should return an array with all Ip Mac adress pair', function() {
```
yyscamper: typo: adress => "address"

lib/services/lookup.js
sunnyqianzhang: Many thanks Xiaofei, very very great comments

sunnyqianzhang:
test this please

lib/services/lookup.js
```diff
@@ -482,6 +482,26 @@ function lookupServiceFactory(
...
+     * objects containing IP Addresses and Mac Addresses associated with the Node.
+     */
+    LookupService.prototype.findIpMacAddresses = function (nodeId) {
+        var self = this;
+
+        return self.validateArpCache()
+        .then(function() {
+            return waterline.lookups.find({node: nodeId}).then(function (records) {
+                return _.reduce(records, function (ipMacAddresses, record) {
+		     ipMacAddresses.push(_.pick(record, ['ipAddress', 'macAddress']));
```
yyscamper: indent problem. I strongly suggest you checking your editor configuration, since there are usually a lot indent problem in your recent PRs.

yyscamper:
@sunnyqianzhang : I believe rebasing your PR will make the jenkins happy

pengz1:
+1

PR #262, opened 2017-03-02
fix random unit-test error due to Node ID Cache in lookup service

yyscamper:
Recently, while I monitoring the Continuous Unit-Test result, I found there was a on-core unit-test error  pattern occurs one in a while, and I tried to stress the on-core unit-test in my local VM, I can reproduced this issue after running enough times (tens of rounds). This is one of the Jenkins job: http://rackhdci.lss.emc.com/job/Continuous-Test/job/Unit-Tests/job/on-core/29836/

I find the root cause is about a tricky async timing control for Promise in unit-test. With my fix in, I have tested the on-core unit-test for more than 100 times, no error happens again.

@anhou @brianparry @iceiilin @pengz1 @lanchongyizu @sunnyqianzhang

### DIFF
--- a/spec/lib/services/lookup-spec.js
+++ b/spec/lib/services/lookup-spec.js
@@ -9,6 +9,7 @@ describe('Lookup Service', function () {
         Errors,
         WaterlineService,
         ChildProcess,
+        Promise,
         sandbox = sinon.sandbox.create();
 
     var lookup = [{
@@ -55,6 +56,7 @@ describe('Lookup Service', function () {
         WaterlineService = helper.injector.get('Services.Waterline');
         MacAddress = helper.injector.get('MacAddress');
         Errors = helper.injector.get('Errors');
+        Promise = helper.injector.get('Promise');
 
         // Mock out the waterline collection methods and initialize them
         var config = {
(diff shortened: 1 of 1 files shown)


spec/lib/services/lookup-spec.js
```diff
@@ -100,17 +101,16 @@ describe('Lookup Service', function () {
...
 
-        it('should resolve pending cache checks once a value is assigned', function (done) {
+        it('should resolve pending cache checks once a value is assigned', function () {
             lookupService.assignNodeIdCache('testAddress', 'nodeId');
-            setTimeout(function () {
-                expect(spy1.called).to.be.ok;
-                expect(spy2.called).to.be.ok;
-                done();
-            }, 0);
+            return Promise.all([
```
anhou: it seems that this unit test case depends on the input of the one above, I think it's better that this one is independent, although maybe it has a little duplication,

yyscamper: Actually, the whole unit-test was written in the dependency way you mentioned, and even worse, there is a large number of other unit-tests are also written in the same way. To break the dependency between each test case is not the intention of my PR, such refactor also need a lot time and code change.

Luckily the mocha runs the test cases in sequence, so there is no problem for such kind of design.

yyscamper:
jenkins: test this please

spec/lib/services/lookup-spec.js
anhou: Although it could run in sequence, and it could pass. it's  not an essential for a pure unit test, the environment should be setup and cleared in setup(), teardown(). otherwise, it's not easy to read, maintain, debug and trace back.  not forcing to change them in this PR.

pengz1:
+1

PR #273, opened 2017-04-18
replace keyword "$" with "___" in mongo to avoid conflict

iceiilin:
Fix https://rackhd.atlassian.net/browse/RAC-4643 "Current RackHD is not compatible with mongodb v2.6.10"

It is caused by conflict keyword "$ref" in both mongoDB and task schema. It is probably that mongodb added compulsory check in version >= 2.6.10, which uncovered this problem.

Since task schema is added to mongoDB when loading task defenition and running a workflow, this fix transforms keyword beginning with "$" to "___" when adding documents in "graphobjects" and "taskdefinition" collection, and recover it after being retrieved from mongodb for task schema validation.

I only add this replacement in these three specific points to minimize the negative impact on performance for recursively tracing "$" inside a JSON data. Besides, restricting this operation in mongo.js to not influence   the business logic of taskgraph, neither other database that rackhd would adopt and doesn't have this conflict.

@yyscamper @brianparry @nortonluo @lanchongyizu

### DIFF
--- a/lib/workflow/stores/mongo.js
+++ b/lib/workflow/stores/mongo.js
@@ -226,6 +226,8 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             new: true,
             upsert: true
         };
+
+        exports._replaceKeyword(definition, '$', '___');
         return waterline.taskdefinitions.findAndModifyMongo(query, {}, definition, options);
     };
 
@@ -292,6 +294,8 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
                 instanceId: 1
             }
         };
+
+        exports._replaceKeyword(graph, '$', '___');
         return waterline.graphobjects.findAndModifyMongo(query, {}, graph, options);
     };
 
@@ -358,6 +362,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             if (_.isEmpty(graph)) {
                 return undefined;
             } else {
+                exports._replaceKeyword(graph, '___', '$');
                 return {
                     graphId: graph.instanceId,
                     context: graph.context,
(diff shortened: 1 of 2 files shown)


lib/workflow/stores/mongo.js
```diff
@@ -1059,5 +1064,32 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
...
+        _.map(data, function(value, key){
+
+            if ((typeof key === "string") && (key.startsWith(beginSrc))) {
+                var newKey = beginDst + key.slice(beginSrc.length);
+                data[newKey] = value;
+                delete data[key];
+                key = newKey;
+            }
+
+            if (typeof value === "object") {
```
yyscamper: How about if `value` is an array?

yyscamper: Sometimes, the $ref may be included in an array, for example:
```javascript
allOf:[
    { $ref: "#/defintions/foo" },
    { $ref: "#/definitions/bar" }
]
```

lib/workflow/stores/mongo.js
```diff
@@ -1059,5 +1064,32 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
...
 
+    /**
+     * Replace document keywords which start with one string
+     * This is to avoid conflict of keywords between RackHD document and mongodb
+     * @param {Object} data - the document that needs to be transformed
+     * @param {String} beginSrc - the key that starts with this string will be replaced
+     * @param {String} beginDst - the key that will be used in the data base
+     * @memberOf store
+     */
+    exports._replaceKeyword = function(data, beginSrc, beginDst) {
```
yyscamper: Actually, IMO, the `beginSrc` and `beginDst` isn't so clear enough. How about below?
```javascript
exports._replaceKeyPrefix = function (data, oldPrefix, newPrefix)
```

spec/lib/workflow/stores/mongo-spec.js
```diff
@@ -239,6 +239,36 @@ describe('Task Graph mongo store interface', function () {
         });
     });
 
+    it('persistTaskDefinition should recursively replace keywords if any', function() {
+        var definition = {
+            injectableName: 'testname',
+            optionsSchema: {
+                version: {
+                    $ref: "abc.json"
```
yyscamper: In unit-test, please add some more complex example to test the recursive works well

spec/lib/workflow/stores/mongo-spec.js
```diff
@@ -333,6 +363,32 @@ describe('Task Graph mongo store interface', function () {
...
+            taskId: task1,
+            graphId: uuid.v4()
+        };
+
+        var graph = {
+            instanceId: 'test',
+            context: {},
+            tasks: {}
+        };
+        graph.tasks[task1] = {___ref: 1};
```
yyscamper: Please add some more complex example to validate the recursive works well.

spec/lib/workflow/stores/mongo-spec.js
```diff
@@ -239,6 +239,36 @@ describe('Task Graph mongo store interface', function () {
         });
     });
 
+    it('persistTaskDefinition should recursively replace keywords if any', function() {
```
yyscamper: You forget to test the `persistGraphObject`

lib/workflow/stores/mongo.js
```diff
@@ -358,6 +362,7 @@ function mongoStoreFactory(waterline, Promise, Constants, Errors, assert, _) {
             if (_.isEmpty(graph)) {
                 return undefined;
             } else {
+                exports._replaceKeyword(graph, '___', '$');
```
yyscamper: The function `getTaskDefinition` should also replace the `___` with `$`

iceiilin:
close this PR. complete solution is still under discussion.

PR #281, opened 2017-06-08
 Add protocol subscribeSelEvents

pengz1:
Add protocol subscribeSelEvents.

subscribeSelEvents can be used to monitor specified SEL serial.

### DIFF
--- a/lib/protocol/events.js
+++ b/lib/protocol/events.js
@@ -420,6 +420,18 @@ function eventsProtocolFactory (
         return self.publishGraphEvent(graphId, 'progress.updated', data.nodeId, data);
     };
 
+    EventsProtocol.prototype.subscribeSelEvent = function (pollerId, nodeId, callback) {
+        assert.isMongoId(pollerId, 'subscribeSelEvent requires pollerId');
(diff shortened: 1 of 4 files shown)


lib/protocol/events.js
```diff
@@ -420,6 +420,18 @@ function eventsProtocolFactory (
...
     };
 
+    EventsProtocol.prototype.subscribeSelEvent = function (pollerId, nodeId, callback) {
+        assert.isMongoId(pollerId, 'subscribeSelEvent requires pollerId');
+        assert.isMongoId(nodeId, 'subscribeSelEvents requires nodeId');
+        assert.func(callback, 'subscribeSelEvents requires callback function');
+
+        return messenger.subscribe(
+            Constants.Protocol.Exchanges.Events.Name,
+            'polleralert.sel.updated.critical.' + pollerId + '.' + nodeId,
```
iceiilin: Is it possible to make severity ("critical" here) also a parameter passed by callers? That might be more flexible for different user scenarios.

pengz1: @iceiilin Sel events here is a series, not one event. If I let user to defined this level, I can't guarantee what I got.

PR #282, opened 2017-06-14
Modifying name and location of smi config json

michaelsteven:
Modifying name and location of smi config json used by the smi discovery and dell wsman inventory microservices.  

This serves two purposes:
   1) removes wsman from the name 
   2) pulls from a path that will be in both ansible and docker installations. 

This change should only affect the taskgraphs that invoke the SMI microservices.

Another PR to physically add the file to the monorail folder via the RackHD/RackHD repo should be committed around the same time.


michaelsteven:
This is the related PR in RackHD repository:
https://github.com/RackHD/RackHD/pull/798

PR #287, opened 2017-07-27
clean up unit-test errors when it timeouts

lanchongyizu:
**Backgroud**
This is for the story https://rackhd.atlassian.net/browse/RAC-5639.

During work of https://rackhd.atlassian.net/browse/RAC-5593, we found many unit test cases won't do the clean-up work if running into failure (aka exception handler) , so the plan is to add clean-up work in unit test cases.

**TEST**
It's tested in sandbox:
http://10.62.59.175:8080/job/on-core_jamie/

@RackHD/corecommitters

### DIFF
--- a/spec/lib/common/arp-spec.js
+++ b/spec/lib/common/arp-spec.js
@@ -22,16 +22,13 @@ describe('ARPCache', function () {
     before(function () {
         arpCache = helper.injector.get('ARPCache');
         fs = helper.injector.get('fs');
-        sinon.stub(fs, 'readFileAsync');
     });
 
     beforeEach(function() {
-        fs.readFileAsync.reset();
+        this.sandbox.stub(fs, 'readFileAsync');
     });
 
-    helper.after(function () {
-        fs.readFileAsync.restore();
-    });
+    helper.after();
 
     describe("Handle ARP Entry", function(){
         it('should parse ARP data', function() {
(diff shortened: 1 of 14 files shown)


anhou:
@lanchongyizu Could you help post the errors in the RAC or in this PR's message? I see lots of minor changes which is not related to exception handler.

lib/services/waterline.js
```diff
@@ -125,8 +129,15 @@ function WaterlineServiceFactory(
         .then(function() {
             //it's OK to create indexes multiple times, so the code isn't protected by initialized
             //checking.
+            //
+            //Catching errors is because:
+            //  (1) When it fails to create indexes, the service still works.
```
anhou: So does it mean that if the error is catched, and the service could continue to work?

spec/lib/common/arp-spec.js
```diff
@@ -22,16 +22,13 @@ describe('ARPCache', function () {
     before(function () {
         arpCache = helper.injector.get('ARPCache');
         fs = helper.injector.get('fs');
-        sinon.stub(fs, 'readFileAsync');
     });
 
     beforeEach(function() {
-        fs.readFileAsync.reset();
+        this.sandbox.stub(fs, 'readFileAsync');
```
anhou: I see that readFileAsync 's stub, restore, reset handling is not aligned in this PR

lanchongyizu:
The main errors are in SH sandbox Jenkins, such as http://10.62.59.175:8080/blue/organizations/jenkins/on-core_jamie/detail/on-core_jamie/35/tests

lib/services/waterline.js
lanchongyizu: Right.

PR #288, opened 2017-08-02
Resolve mongodb compatibility problem

anhou:
Fix unittest error like 'MongoError: Index with name: identifier_1 already exists with different options' in mongodb 2.6.x/3.x. It's a mongodb compatibility problem between 2.4.x and 2.6.x/3.x.  

new version prevent sails-mongo creating duplicated index operations. either 'primaryKey:true' or 'unique:true' would create index, so following creating index opertaion would be duplicated, so the solution is to remove duplicated create index operation.

I've checked that the indexes are the same before and after this PR, they're as below. there's no impacts.
```
> db.environment.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.environment"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "identifier" : 1
                },
                "name" : "identifier_1",
                "ns" : "pxe.environment"
        }
]
> db.files.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.files"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "basename" : 1
                },
                "name" : "basename_1",
                "ns" : "pxe.files"
        }
]
> db.graphobjects.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.graphobjects"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "instanceId" : 1
                },
                "name" : "instanceId_1",
                "ns" : "pxe.graphobjects"
        }
]
> db.localusers.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.localusers"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "username" : 1
                },
                "name" : "username_1",
                "ns" : "pxe.localusers"
        }
]
> db.lookups.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.lookups"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "macAddress" : 1
                },
                "name" : "macAddress_1",
                "ns" : "pxe.lookups"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "macAddress" : 1,
                        "ipAddress" : 1
                },
                "name" : "macAddress_1_ipAddress_1",
                "ns" : "pxe.lookups"
        },
        {
                "v" : 1,
                "key" : {
                        "node" : 1
                },
                "name" : "node_1",
                "ns" : "pxe.lookups"
        },
        {
                "v" : 1,
                "key" : {
                        "ipAddress" : 1
                },
                "name" : "ipAddress_1",
                "ns" : "pxe.lookups"
        }
]
> db.roles.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.roles"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "role" : 1
                },
                "name" : "role_1",
                "ns" : "pxe.roles"
        }
]

> db.taskdefinitions.getIndexes()
[
        {
                "v" : 1,
                "key" : {
                        "_id" : 1
                },
                "name" : "_id_",
                "ns" : "pxe.taskdefinitions"
        },
        {
                "v" : 1,
                "unique" : true,
                "key" : {
                        "injectableName" : 1
                },
                "name" : "injectableName_1",
                "ns" : "pxe.taskdefinitions"
        }
]
>

```

### DIFF
--- a/lib/models/graph-object.js
+++ b/lib/models/graph-object.js
@@ -56,12 +56,6 @@ function GraphModelFactory (
                 var obj = this.toObject();
                 return Constants.Task.ActiveStates.indexOf(obj._status) > -1;
             }
-        },
-        $indexes: [
-            {
-                keys: { instanceId: 1 },
-                options: { unique: true }
-            }
-        ]
+        }
     });
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ services:
  - mongodb
  - rabbitmq
 
+script:
+ - mongo --version
+ - npm test
+
 env:
   global:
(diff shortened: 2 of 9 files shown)


lib/models/graph-object.js
```diff
@@ -56,12 +56,6 @@ function GraphModelFactory (
                 var obj = this.toObject();
                 return Constants.Task.ActiveStates.indexOf(obj._status) > -1;
             }
-        },
-        $indexes: [
-            {
-                keys: { instanceId: 1 },
```
yyscamper: previously, every index is carefully designed to satisfy the frequently database query, so if index is removed, the performance will certainly be impacted. Usually, you could only see the performance issue during scale environment. So I'd suggest you doing some scale testing before removing these indexes.

From the implementation of waterline mongo adapter, see if the "unique" of one attribute is set as true, then waterline will automatically create an unique index for it (see code https://github.com/balderdashy/sails-mongo/blob/7572dd47a1e4b523d728e6ad4cd30bdf6b4174fc/lib/index.js#L768). So this may be the source of duplicated index, we could remove those index setting if the attribute is set as "true", or we can always remove the unique from attribute and keep using our own index setting.

anhou: @yyscamper that's why I remove some indexes, they're duplicated, and I could check the indexes by db.xx.getIndexes() to see if the database index are affected instead of scaling test.

actually, not only 'unique: true' would affect the index, but also 'primaryKey: true' so I prefer attributes keep the same with before, and check if new added index is duplicated, if yes, remove them, if not, leave there, to reduce the impacts.

anhou:
test this please

PR #289, opened 2017-08-02
RAC-5804 - Change default config messages to be info

jimturnquist:
log message that reports default config settings is
being made level info instead of warning.

'Configuration value is undefined, using default'


jgolio:
LGTM

PR #291, opened 2017-08-09
add graphId and graphName for event of graphStarted and graphFinished

lanchongyizu:
1) update eventsProtocol.publishGraphFinished
2) update eventsProtocol.publishGraphStarted

@RackHD/corecommitters

### DIFF
--- a/lib/protocol/events.js
+++ b/lib/protocol/events.js
@@ -203,12 +203,13 @@ function eventsProtocolFactory (
     };
 
 
-    EventsProtocol.prototype.publishGraphStarted = function (graphid, status, nodeId) {
+    EventsProtocol.prototype.publishGraphStarted = function (graphid, data, nodeId) {
         var self = this;
         assert.uuid(graphid);
-        assert.string(status);
-
-        var data = { status: status };
+        assert.object(data);
+        assert.uuid(data.graphId);
+        assert.string(data.graphName);
+        assert.string(data.status);
 
         return messenger.publish(
             Constants.Protocol.Exchanges.Events.Name,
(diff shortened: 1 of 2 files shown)


lanchongyizu:
Jenkins: depends on https://github.com/RackHD/on-taskgraph/pull/269
test this please.

lib/protocol/events.js
```diff
@@ -203,12 +203,13 @@ function eventsProtocolFactory (
...
 
-    EventsProtocol.prototype.publishGraphStarted = function (graphid, status, nodeId) {
+    EventsProtocol.prototype.publishGraphStarted = function (graphid, data, nodeId) {
         var self = this;
         assert.uuid(graphid);
-        assert.string(status);
-
-        var data = { status: status };
+        assert.object(data);
+        assert.uuid(data.graphId);
```
anhou: Don't need to check data's members here. it's adding limitation to this funciton, you could add checking in consuming place

lib/protocol/events.js
```diff
@@ -227,17 +228,18 @@ function eventsProtocolFactory (
...
 
-    EventsProtocol.prototype.publishGraphFinished = function (graphid, status, nodeId) {
+    EventsProtocol.prototype.publishGraphFinished = function (graphid, data, nodeId) {
         var self = this;
         assert.uuid(graphid);
-        assert.string(status);
-
-        var data = { status: status };
+        assert.object(data);
+        assert.uuid(data.graphId);
```
anhou: Don't need to check data's members here. it's adding limitation to this function, you could add checking in consuming place

lib/protocol/events.js
lanchongyizu: For this function, we should check `graphId, graphName and status` to keep the correct usage.

lib/protocol/events.js
lanchongyizu: For this function, we should check `graphId, graphName and status` to keep the correct usage.

lib/protocol/events.js
anhou: I'm not sure what's the correct usage, I think it's  displayed information, so I think your purpose is not to protect this function, because not checking wouldn't affect this function, if you just want to make sure information display correctly, add the type check in calling method or callback method

lanchongyizu: The correct usage is that when we call the function, we should provide `graphId, graphName and status`.

lanchongyizu: So the function `publishExternalEvent ` you commited before(https://github.com/RackHD/on-core/blob/master/lib/protocol/events.js#L312-L318)  should either add check function or don't check the data to let it more tolerance.
I think we should keep the format aligned here.

lib/protocol/events.js
anhou: Yes, it's also better to add a function to check the data. not so good before

PR #292, opened 2017-08-09
Make aggregated log column align and remove redundant field in logs

sunnyqianzhang:
To fix this issue:
https://rackhd.atlassian.net/browse/RAC-5515

What the code change did:
1. Make the severity level the same length in logs. e.g. [  info  ], [critical], they have the same length
2. Remove the fields like [on-taskgraph][ TaskGraph.TaskScheduler], as they don't provide very useful information but makes the logs complicated.

Before code change, the log is like this:
 0|taskgrap | [debug] [2017-08-09T08:53:30.463Z][on-taskgraph][ TaskGraph.TaskScheduler][ Server] Running job.

After the code change, the log is like:
0|taskgrap | [ debug  ] [2017-08-09T08:53:30.463Z] [ Server]Running job.

@anhou @pengz1 @iceiilin

### DIFF
--- a/lib/serializables/log-event.js
+++ b/lib/serializables/log-event.js
@@ -51,9 +51,6 @@ function LogEventFactory (
             host: {
                 type: 'string'
             },
-            module: {
-                type: 'string'
-            },
             level: {
                 type: 'string',
                 enum: _.keys(Constants.Logging.Levels)
@@ -74,7 +71,7 @@ function LogEventFactory (
                 type: 'string'
             }
         },
-        required: [ 'module', 'level', 'timestamp', 'caller', 'subject' ]
+        required: [ 'level', 'timestamp', 'caller', 'subject' ]
     };
 
     Serializable.register(LogEventFactory, LogEvent);
@@ -83,10 +80,11 @@ function LogEventFactory (
         var statement = [];
 
         this.context = LogEvent.redact(this.context);
-        statement.push('[%s]'.format(this.level));
+        // The level length 8 is hardcoded here, as the loggest level up to now
+        // is 8, this might be subject to change when new log levels are added
+        // in the future
+        statement.push('[%s]'.format( _.pad(this.level, 8)));
         statement.push('[%s]'.format(this.timestamp));
-        statement.push('[%s]'.format(this.name));
-        statement.push('[%s]'.format(this.module));
         statement.push('[%s]'.format(
             this.subject.substring(this.subject.length - Constants.Logging.Context.Length)
         ));


lib/serializables/log-event.js
```diff
@@ -51,9 +51,6 @@ function LogEventFactory (
             host: {
                 type: 'string'
             },
-            module: {
```
yyscamper: I think the `module` does provides useful information, as it quickly tells you which module generates the message and so you can quickly go to the source code.

Suppose if there is a log message "running xxxx", if the module is removed from logging entry, then you need to search the whole code repositories to find out which module generates this message, if there are multiple modules generate same message, then you will be lost and you need additional information to locate which module it is.

yyscamper:
if you want to help the log analysis, I recommend you using the structured logging, see https://stackify.com/what-is-structured-logging-and-why-developers-need-it/ .

In node.js, a very popular structure logging library is `bun-yan` (https://github.com/trentm/node-bunyan), it outputs every log in JSON format, and also provides a cli tool to convert the JSON to more human-readable format.

lib/serializables/log-event.js
anhou: The module name is duplicated with the file name, file name could give us the same information, and the module name is the same with file name it's just for developer, not for normal users. or another way is add module name to file name line, although it's duplicated, it could be quick map for module and file

lib/serializables/log-event.js
```diff
@@ -83,13 +77,11 @@ function LogEventFactory (
...
         this.context = LogEvent.redact(this.context);
-        statement.push('[%s]'.format(this.level));
+        // The level length 8 is hardcoded here, as the loggest level up to now
+        // is 8, this might be subject to change when new log levels are added
+        // in the future
+        statement.push('[%s]'.format( _.pad(this.level, 8)));
         statement.push('[%s]'.format(this.timestamp));
-        statement.push('[%s]'.format(this.name));
-        statement.push('[%s]'.format(this.module));
-        statement.push('[%s]'.format(
```
anhou: we still need [Server], when having nodeId, it will display [<nodeId>] such as [8d2f33]

sunnyqianzhang: Ok, thanks andrew

lib/serializables/log-event.js
sunnyqianzhang: @yyscamper xiaofeifei, At first I had the same thought as you did. Then when I looked deeper into the logs, I found the log anyway will give the file name and line number. So removing module names won't affect the analysis, but makes the log less.
1|http     | [ debug  ] [2017-08-09T19:23:30.688Z] advertise-alive
1|http     |  -> /lib/services/upnp-service.js:103

sunnyqianzhang: @yyscamper Ok, Let me see see the bun-yan

lanchongyizu: @sunnyqianzhang @yyscamper I also think we should keep the field `module`, which provides more accurate info.

PR #293, opened 2017-08-16
RAC5843: enhance HWIMO-TEST to handle the dependencies

PengTian0:
The unit test depends on mongodb and rabbitmq.
The PR enhances the script HWIMO-TEST to run unit test with checking and setup dependencies.
@anhou @iceiilin @lanchongyizu @panpan0000

### DIFF
--- a/HWIMO-TEST
+++ b/HWIMO-TEST
@@ -1,21 +1,69 @@
-#!/bin/bash
-
+#!/bin/bash -e
 # Copyright 2015, EMC, Inc.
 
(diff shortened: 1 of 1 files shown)


PengTian0:
test this please

lanchongyizu:
Since the unit test still depends on DNS, which requires `127.0.0.1 $hostname` in /etc/hosts, so it's better to add the checking. The same problem exists in https://github.com/RackHD/on-http/pull/736

PR #295, opened 2017-08-23
Add concourse testing scripts

keedya:


keedya:
@johren @geoff-reid @hohene

PR #317, opened 2017-11-06
Unit test fix for mocha update

mcgG:
RAC:  [https://rackhd.atlassian.net/browse/RAC-5743](https://rackhd.atlassian.net/browse/RAC-5743)
----
Paired with @PengTian0 

Using both `return promise` and `done()` in async unit test case will cause following error:
`Resolution method is overspecified. Specify a callback *or* return a Promise; not both.`

According to Mocha's [official site](http://mochajs.org/):
`In Mocha v3.0.0 and newer, returning a Promise and calling done() will result in an exception`

**Solution**
----
Only call done() instead of return a promise in all async unit test cases.

@PengTian0 @pengz1 @anhou @iceiilin @bbcyyb

### DIFF
--- a/spec/lib/common/child-process-spec.js
+++ b/spec/lib/common/child-process-spec.js
@@ -187,7 +187,7 @@ describe("ChildProcess", function () {
 
             // Use done() because if the promise is fulfilled that is a failure
             // case for this test. Assert that we hit this catch block.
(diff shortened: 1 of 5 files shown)


bbcyyb:
+1

PR #318, opened 2017-11-08
RAC-5743: unit test doesn't exit after upgrading mocha to 4.0.1

PengTian0:
Paired with @mcgG 
Unit test doesn't exit after upgrading mocha to 4.0.1 because the new version of mocha won't exit if there are running process.
There are 3 kinds of cases that make mocha hangs
1. amqp: try to reconnect after disconnect(https://github.com/postwait/node-amqp/issues/462).
    Fix: set the option "reconnect" to false in stop() to avoid the issue.
2. statsd: the socket is not closed after call statsd.stop()
    Fix: call close() in stop() to close socket.
3. stub object messenger without restore. That leads to the messenger.stop() failed to stop amqp connection.
    Fix: replace stub object with stub method of object in beforeEach. The stubbed methods will be restored in afterEach().

@mcgG @anhou @pengz1

### DIFF
--- a/lib/common/connection.js
+++ b/lib/common/connection.js
@@ -102,7 +102,7 @@ function connectionFactory (assert, amqp, Promise, util, _) {
                     delete self.connection;
                     resolve();
                 });
-
+                self.connection.setImplOptions({reconnect: false});
                 self.connection.disconnect();
             } else {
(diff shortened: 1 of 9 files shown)


mcgG:
👍 +1

PR #322, opened 2018-02-17
Replace shortid to nanoid

ai:
I replaced ID generator because of:

1. `shortid` has no maintenance anymore.
2. `shortid` has some dangerous issues in tracker like bad distribution (higher collision probability).
3. [`nanoid`](https://github.com/ai/nanoid) is safer and 9 times faster than `shortid`
4. There were no changes for project users. Only a little better performance and smaller `node_modules`. You will have safer ID generator with better support.

@anhou @jimturnquist what do you think about it?


ai:
@jimturnquist do you have any questions?

PR #324, opened 2018-03-01
upgrade node and npm in Dockerfile

lanchongyizu:
**Background**

Node images in `nodesource `aren't updated for almost one year, and we need to upgrade `node v4` to be compatible with `npm5`.

**Done**

* upgrade node and npm in Dockerfile

@RackHD/corecommitters @pengz1 @PengTian0


lanchongyizu:
@keedya It's just to keep aligned with the original debian release, and newer release has bigger size of images. I am fine with moving to a newer debian  release.  How about jessie? 

And this is not part of upgrading to node 6, but for package lockdown with `npm5 `which requires `node v4.7.0+`. Please refer https://github.com/RackHD/on-tasks/pull/585.

We have other stories to enable concourse pipelines to test rackhd with node v4, v6 and v8.

lanchongyizu:
We have merged the code in https://github.com/RackHD/on-core/pull/325, and plan to close the PR.

PR #329, opened 2018-05-08
redact productkey

lanchongyizu:
When user install windows 2012, the logs include `productkey:      xxxxx-xxxxx-xxxxx-xxxxx`, which is private and should be redacted.

@RackHD/corecommitters @nortonluo @iceiilin


lkx007:
+1

